PR #141, opened 2020-02-21
Upgrade deno to v0.34.0

player-4:


src/http/server.ts
```diff
@@ -58,9 +58,9 @@ export default class Server {
    *     A property to hold the location of this server on the filesystem. This
    *     property is used when resolving static paths.
    *
-   * @property string directory
+   * @property null|string directory
```
crookse: Undefined

src/http/server.ts
```diff
@@ -164,7 +164,11 @@ export default class Server {
     request = await Drash.Services.HttpRequestService.hydrate(request, {
       default_response_content_type: this.configs.response_output,
       memory_allocation: {
-        multipart_form_data: this.configs.memory_allocation.multipart_form_data
```
crookse: I get it, but clean up

src/http/server.ts
```diff
@@ -263,8 +270,8 @@ export default class Server {
   public handleHttpRequestError(
     request: any,
     error: any,
-    resource: Drash.Http.Resource = null,
```
crookse: Add comment stating why we use any here

src/services/http_request_service.ts
```diff
@@ -249,22 +252,31 @@ export default class HttpRequestService {
...
       request,
-      options.default_response_content_type
+      contentType
     );
 
     // Parse the body now so that callers don't have to use async-await when
     // trying to get the body at a later time. We're sacrificing performance for
     // convenience here.
-    const pb: ParsedRequestBody = await this.parseBody(request, options);
+    const pb: ParsedRequestBody = await this.parseBody(
```
crookse: should move the ternary to the method

src/services/http_request_service.ts
```diff
@@ -335,17 +344,29 @@ export default class HttpRequestService {
...
+        if (match) {
+          boundary = match[1];
+        }
+        if (!boundary) {
+          return ret;
+        }
       } catch (error) {
         throw new Error(`Error trying to find boundary.\n` + error.stack);
       }
+      if (!boundary) {
```
crookse: You’re already checking for this in the try block

src/services/http_request_service.ts
```diff
@@ -335,17 +344,29 @@ export default class HttpRequestService {
...
+        return ret;
+      }
       try {
+        const maxMemory = options && options.memory_allocation && options.memory_allocation.multipart_form_data
```
crookse: We’re checking twice to set 10MB? I saw this up above. Should come back around to this to see how to clean it up

console/typescript/docs_compiler.ts
```diff
@@ -7,8 +7,8 @@ import { BufReader } from "../../deps.ts";
 
 export default class ClassCompiler {
 
-  protected path: string;
-  protected process: string;
+  protected path: string = "";
```
saragee3: why initialize this with an empty string?

player-4: i ran into `Property 'path' has no initializer and is not definitely assigned in the constructor.`. i changed the logic: 12c96d79ad7ad116558ee247b45dbd0009924d0e

crookse: @saragee3 can you finish your review please?

deps.ts
```diff
@@ -1,40 +1,35 @@
 export {
   serve,
   ServerRequest
-} from "https://deno.land/std@v0.32.0/http/server.ts";
+} from "https://deno.land/std@v0.34.0/http/server.ts";
```
saragee3:
```
const DENO_STD_URL = `https://deno.land/std@v0.34.0/`

export {
 serve,
 ServerRequest
} from `{DENO_STD_URL}http/server.ts`;
```

Just a thought.... might start to get cumbersome if this file gets bigger. Maybe this is something that we are already thinking about adding to the minimal imports feat... in that case, would be good to have a prop for this deno_std_url?

crookse: That sounds good, but I'm pretty sure that would require dynamic imports since variables can't go into the URL. I could add in the `bump_versions` script again. It took a `from` and a `to` variable and would update the versions `from` #.#.# `to` #.#.#. I'll make an issue about it though.

Issue #143, opened 2020-02-21
Cookie handling


crookse:
Please view https://github.com/drashland/deno-drash/blob/master/DEV.md before starting.

## Summary

What: Handle cookies.

Why: Cookies are part of the request-resource-response lifecycle, but the Drash server doesn't handle them yet.

## Acceptance Criteria

- [x] documentation (should be under the Misc section and it should reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
- [x] tests
- [x] ~cookie schema matches schema in MDN~
- [x] use `Cookie` interface that Deno has defined
- [x] ~resources should be able to call `this.server.getCookie()` and `this.server.setCookie()`.~
- [x] resources should be able to get cookies from the request. the `HttpRequestService` should be the class that handles parsing cookies; and resources should be able to call `this.request.getCookie()` or something similar that makes sense
- [x] resources should be able to set cookies on the response (e.g., `this.response.setCookie()` or something similar that makes sense)
- [x] ~build `api_reference.json` file and update `deno-drash-docs` to use it.~

## Example Pseudo Code (for implementation)

```typescript
server.setCookie();
server.getCookie();
```
The methods above should implement https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies as closely as possible

ebebbington:
I was thinking of taking this on and came to a conclusion on how this could be implemented:

The `this.request.headers.*` is always undfined so I can't try the `cookie` property (not sure how it's accessed in Drash, assuming there's another way but i'm testing in my own application that uses Drash as the Server), but the cookies come in look like a string e.g.: `cookiename=cookieval; cookiename2=cookieval2`, so assuming this, we could create a `cookies` property on the `Request` or `Server` class:
```
public cookies: [{name: string, value: string}]
```

Very early in the lifecycle, we can extract the cookies from the request like so:
```
public extractRequestCookies () {
 const splitCookies = this.request.headers.cookie.split(';')
 splitCookies.forEach(cookie => {
    this.request.cookies.push({name: cookie.split('=')[0], value: cookie.split('=')[1]
 })
}
```

Now we can create out getter and setter:

```
public setCookie({name: string, value: string}): void {
  this.request.cookies.push({name: name, value: value);
  // fixme :: will actually need to set it in the `headers` property
}
public getCookie (cookieName: string): string|boolean {
  let foundCokie ;
  this.request.cookies.forEach((cookie: Cookie) => {
[...]

PR #146, opened 2020-02-24
Added Cookie handling

ebebbington:
Implements #143 

I will make a pr to deno-drash docs once this PR is approved in case someone wants me to adjust something which could lead to a doc block change.

**Developments**
Added the ability to get a cookie, set a cookie and delete a cookie. This functionality is about 95% Deno, utilising their cookie module. It conforms to the standards e.g. include `path`, `expiry` etc.

The get cookie function resides inside the request object, and will use Deno's `getCookies` to retrieve a single cookie, as there isn't an already built function to get a single one.

The delete cookie and set cookie functions reside inside the response class. The reason why these functions are re-defined is so it can allow type hints whilst writing the code as simply doing `public setCookie = setCookie` threw no errors when passing in 0 params, so they were changed to the following:
```
// so it can be strict
public setCookie (cookie: Cookie) { setCookie(this, cookie) }
```

**How to Use**
To get cookies:
`const carCookieValue = this.request.getCookie('car')`

To set cookies:
```
const cookie: Cookie = {
  name: 'car',
  value: 'Ford',
  expiry: 'some date'
  ...
}
this.response.setCookie(cookie)
```

To delete a cookie:
`this.response.delCookie('car')`

**Resources**
Deno's cookie module (holds interface for Cookie as well): https://deno.land/std/http/cookie.ts

### DIFF
--- a/deps.ts
+++ b/deps.ts
@@ -33,3 +33,10 @@ export {
 export {
   MultipartReader, FormFile
 } from "https://deno.land/std@v0.34.0/mime/multipart.ts";
+
+export { 
+  getCookies,
+  setCookie,
+  delCookie,
+  Cookie
+} from "https://deno.land/std/http/cookie.ts"
--- a/deps.ts
+++ b/deps.ts
@@ -39,4 +39,4 @@ export {
   setCookie,
   delCookie,
   Cookie
-} from "https://deno.land/std/http/cookie.ts"
+} from "https://deno.land/std@v0.34.0/http/cookie.ts"
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -1,5 +1,6 @@
 import Drash from "../../mod.ts";
 import { STATUS_TEXT, Status } from "../../deps.ts";
+import { setCookie, delCookie, Cookie } from "../../deps.ts"
 
 /**
  * @memberof Drash.Http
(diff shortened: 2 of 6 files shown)


ebebbington:
Closed as I had forgotten to write the tests... Woops... Will reopen once finished

ebebbington:
Actually I don't think it's possible to write tests for it because it isn't actually used yet?

ebebbington:
Any info on what those errors mean?

crookse:
i'm guessing it has something to do with your forked branch not being able to checkout drash. it's fine. we'll see how it does with the release/v0.34.x branch

deps.ts
```diff
@@ -33,3 +33,10 @@ export {
 export {
   MultipartReader, FormFile
 } from "https://deno.land/std@v0.34.0/mime/multipart.ts";
+
+export { 
+  getCookies,
+  setCookie,
+  delCookie,
+  Cookie
+} from "https://deno.land/std/http/cookie.ts"
```
crookse: Please use `v0.34.0` deno std. Hopefully in the near future we can make adding these deps in configurable.

ebebbington:
Tested, all tests pass, also as a side test in the browser checked the dev tools for saving, getting and removing cookies in the resource (these passed as well)

src/http/response.ts
```diff
@@ -1,5 +1,6 @@
 import Drash from "../../mod.ts";
 import { STATUS_TEXT, Status } from "../../deps.ts";
+import { setCookie, delCookie, Cookie } from "../../deps.ts"
```
crookse: since `deps.ts` is already being imported on line 2, you can add `setCookie`, `delCookie`, `Cookie` to line 2. just make sure you stack the list vertically and put it in alphabetical order

src/services/http_request_service.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from "../../deps.ts";
 import StringService from "./string_service.ts";
 import { ParsedRequestBody } from "../interfaces/parsed_request_body.ts";
+import { getCookies, Cookie } from "../../deps.ts"
```
crookse: See comment above

Issue #155, opened 2020-03-24
Deno v0.36.0 and v0.37.0 support


eliassjogreen:
## Summary

This is probably something already being worked on and i am probably just nagging but deno v0.36.0 and v0.37.0 support would be nice to have. Currently when i tried to upgrade the dependencies failing in the old version it failed at the following test: `missing CSRF token` and errored at the following test `wrong CSRF token` (#138). When running the example everything seemed to work fine otherwise.

crookse:
@eliassjogreen,

It’s done!

crookse:
The tests have been problematic since deno v0.34.0. Not sure why. I did make another ticket though to address those issues. Everything does look fine otherwise like you said.

PR #156, opened 2020-03-24
#155 Updated std version to v0.37.0

eliassjogreen:


crookse:
thanks for this @eliassjogreen. merging your branch into `upgrade-deno`

PR #171, opened 2020-04-13
Feature/issue 161

ebebbington:
Abstracted the logic a user needs to do for:

* rendering basic html files

* rendering html files with dynamic data e.g. `{{ name }}`

* rendering Drash's template engine files

into a new method in `Drash.Http.Response`

As a result, i have added the following:

* Two new properties to `Drash.Http.Response`

* Two new properties for `Drash.Interfaces.ServerConfigs

* Created a 2nd param for `Drash.Http.Response` that uses `Drash.Interfaces.ResponseOptions`, for passing in the new properties in `Drash.Http.Server`

* Created tests for the following:

    * Using the render method for returning basic html
    * Using the render method for returning html that has dynamic content
    * Using the render method when `template_engine` is true for server instantiation


src/http/response.ts
```diff
@@ -44,22 +45,84 @@ export default class Response {
...
    * @description
    *     Construct an object of this class.
    *
    * @param any request
+   *
+   * @param ResponseOptions options
+   *     See Drash.Interfaces.ResponseOptions
    */
-  constructor(request: any) {
+  constructor(request: any, options: ResponseOptions = {}) {
```
crookse, line 76: instead of importing `ResponseOptions`, use `Drash.Http.ResponseOptions`

src/http/response.ts
```diff
@@ -44,22 +45,84 @@ export default class Response {
...
+    const filename = this.views_path += args[0];
+    const data = args.length >= 2 ? args[1] : null
+    if (this.template_engine && data) {
+      const engine = new Drash.Compilers.TemplateEngine();
+      return engine.render(filename, data);
+    }
+    const fileContentsRaw = Deno.readFileSync(filename);
+    const decoder = new TextDecoder();
+    let decoded = decoder.decode(fileContentsRaw);
+    if (data) {
```
crookse, line 117: does it change anything if Drash's template engine handles dynamic data?

src/http/response.ts
ebebbington: Addressed this in the below commit `Improved the options parameter interface for Drash.Http.Response`

src/http/response.ts
ebebbington: Addressed this in the below commit `Updated render method to not handle basic dynamic html`

Issue #175, opened 2020-04-15
import { Drash } instead of import Drash


crookse:
## Summary

What:

```
import { Drash } from "https://deno.land/x/drash/mod.ts";
```

Why:

- Drash in its current form doesn't work with `deno doc`.
- To export Drash from a `deps.ts` file, you need to do the following:
    ```
    import Drash from 'https://deno.land/x/drash@v0.39.5/mod.ts';
    export { Drash };
    ```

## Acceptance Criteria

- [x] Drash works with deno doc
- [x] Update documentation
- [x] Update tests

crookse:
For the time being, `export { Drash as default } from "url";` should be used.

PR #178, opened 2020-04-15
Issue #175 support deno doc

crookse:
Addresses #175 

This is a breaking change PR. Drash will now need to be imported like ...

```
import { Drash } from "https://deno.land/x/drash/mod.ts";
```

instead of ...

```
import Drash from "https://deno.land/x/drash/mod.ts";
```

### DIFF
--- a/src/compilers/template_engine.ts
+++ b/src/compilers/template_engine.ts
@@ -2,11 +2,31 @@ const decoder = new TextDecoder();
 
 export default class TemplateEngine {
 
+  /**
+   * @description
+   *     A property to hold the base path to the template.
+   *
+   * @property string views_path
+   */
   public views_path: string = "";
+
+  // FILE MARKER: CONSTRUCTOR //////////////////////////////////////////////////
+
+  /**
+   * @description
+   *     Construct an object of this class.
+   *
+   * @param number code
+   *     The HTTP response code associated with this exception.
+   * @param string message
+   *     (optional) The exception message.
+   */
 
   constructor(viewsPath: string) {
     this.views_path = viewsPath;
   }
+
+  // FILE MARKER: METHODS - PUBLIC /////////////////////////////////////////////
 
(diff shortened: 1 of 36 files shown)


ebebbington:
On the commit `[issue-#175-support-deno-doc] test all of the things` (`
3436713`), because the multiple commands to run the tests aren't running if the previous one succeeds (eg `&&`), if the first test command fails and the second one passed wont this mean (for CI) that the tests have passed?

crookse:
> On the commit `[issue-#175-support-deno-doc] test all of the things` (` 3436713`), because the multiple commands to run the tests aren't running if the previous one succeeds (eg `&&`), if the first test command fails and the second one passed wont this mean (for CI) that the tests have passed?

Good point

ebebbington:
> > On the commit `[issue-#175-support-deno-doc] test all of the things` (` 3436713`), because the multiple commands to run the tests aren't running if the previous one succeeds (eg `&&`), if the first test command fails and the second one passed wont this mean (for CI) that the tests have passed?
> 
> Good point

Could chain them together?
```
run one command \
  && run the other \
  && everything checks out
```
Or whatever way looks better, im sure theres a dozen ways to do this

src/compilers/template_engine.ts
```diff
@@ -2,12 +2,32 @@ const decoder = new TextDecoder();
...
+
+  /**
+   * @description
+   *     Construct an object of this class.
+   *
+   * @param number code
+   *     The HTTP response code associated with this exception.
+   * @param string message
+   *     (optional) The exception message.
+   */
```
ebebbington, line 23: Little picky.. but doesn't seem to be a `code` param passed in?

crookse: LOL copy paste.

crookse: fixed

Issue #173, opened 2020-04-14
Accept vs Response-Content-Type


Guergeiro:
## Discussion
Why use a custom `Response-Content-Type` header instead of the one already defined and standardized ([Accept](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1))?

crookse:
You bring up a very good point actually. So like I said in the YouTube chat, the `Response-Content-Type` header tells the server exactly what content type the user wants. I know `Accept` does the same, but it's not as explicit as they use the words, "We 'prefer' this content type, but any of the following are fine as well."

https://drash.land/docs/#/advanced-tutorials/content-negotiation/user-profiles/part-4 has a bit more on the `Response-Content-Type` header.

But like I said, you bring up a very good point. During my development for serving a specific content type and setting the `Content-Type` header on the response, I realized `Accept` isn't as explicit because it's a list of content types as opposed to just one. Do you think it's ok that the response object uses the `Accept` header and chooses the first content type in the list and sets that as the response's `Content-Type`?

It's really open for discussion. I'm all for using the `Accept` header, but I just don't know how to make it more explicit when it comes to Drash. Any ideas you have are greatly appreciated!

crookse:
This is an example of how the `Accept` header could be used as of right now:

```typescript
import Drash from "https://deno.land/x/drash@v0.39.5/mod.ts";

class HomeResource extends Drash.Http.Resource {

  static paths = ["/"];

  public GET() {
    this.response.body = false;

    let contentTypes = this.request.headers.get("Accept").split(";");

    contentTypes.forEach((contentType: string) => {
      if (!this.response.body) {
        switch (contentType) {
          case "text/html":
            this.response.headers.set("Content-Type", contentType);
            this.response.body = this.htmlResponse();
            break;
          case "application/json":
            this.response.headers.set("Content-Type", contentType);
            this.response.body = this.jsonResponse();
            break;
          case "text/xml":
            this.response.headers.set("Content-Type", contentType);
            this.response.body = this.xmlResponse();
            break;
        }
      }
    });

    if (!this.response.body) {
      this.response.body = "You requested a content-type that isn't supported by this resource.";
    }

    return this.response;
  }

  protected jsonResponse() {
    return {hello: "world"};
  }

  protected htmlResponse() {
    return "<div>Hello, world!</div>";
  }

  protected xmlResponse() {
    return "<body>Hello, world!</body>";
  }
}

[...]

PR #179, opened 2020-04-15
Issue #173 accept content type

ebebbington:
**Overview**
Also removed an old fixme that i had left in whilst i was writing the tests (the fixme was resolved in the leaking-tests issue)

* Added a new `accepts` method

* Tests written

* `api_reference.json` will be updated once this PR is closed and completed

* Wonder if other parts of the docs might need to be updated so developers would use this apporach instead?

### DIFF
--- a/example_app/request_accepts_resource_test.ts
+++ b/example_app/request_accepts_resource_test.ts
@@ -1,0 +1,25 @@
+import members from "../tests/members.ts";
+members.test("RequestAcceptsResource", async () => {
+  let response;
+  let json;
+
+  // Accepts the correct type the resource will give
+  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+    headers: {
+      "Accept": "application/json",
+    },
+  });
+  json = JSON.parse(await response.json());
+  members.assert.equals(json['success'], true);
+  members.assert.equals(json.message, "Responding with the accepted content type");
+
+  // Does not accept the type the resource expects
+  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+    headers: {
+      "Accept": "text/html",
+    },
+  });
+  json = JSON.parse(await response.json());
+  await members.assert.equals(json.success, false);
+  await members.assert.equals(json.message, 'This resource only accepts JSON');
+});
--- a/example_app/request_accepts_resource_test.ts
+++ b/example_app/request_accepts_resource_test.ts
@@ -2,24 +2,47 @@ import members from "../tests/members.ts";
 members.test("RequestAcceptsResource", async () => {
   let response;
   let json;
+  let typeToCheck;
 
-  // Accepts the correct type the resource will give
-  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+  // Accepts the correct type the resource will give - tests calling the `accepts` method with a string and finds a match
+  typeToCheck = "application/json";
+  response = await members.fetch.get("http://localhost:1667/request-accepts?typeToCheck=" + typeToCheck, {
     headers: {
       "Accept": "application/json",
     },
   });
   json = JSON.parse(await response.json());
-  members.assert.equals(json['success'], true);
-  members.assert.equals(json.message, "Responding with the accepted content type");
+  members.assert.equals(json.success, true);
+  members.assert.equals(json.message, "application/json");
 
-  // Does not accept the type the resource expects
-  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+  // Does not accept the type the resource expects - tests calling the `accepts` method with a string with no match
+  response = await members.fetch.get("http://localhost:1667/request-accepts?typeToCheck=" + typeToCheck, {
     headers: {
       "Accept": "text/html",
     },
   });
   json = JSON.parse(await response.json());
   await members.assert.equals(json.success, false);
-  await members.assert.equals(json.message, 'This resource only accepts JSON');
+  members.assert.equals(json.message, undefined);
+
+  // Accepts the first content type - tests when calling the `accepts` method with an array and finds a match
+  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+    headers: {
+      "Accept": "text/xml,text/html,application/json;0.5;something",
+    },
+  });
+  json = JSON.parse(await response.json());
+  members.assert.equals(json.success, true);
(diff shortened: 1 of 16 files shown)


src/services/http_request_service.ts
```diff
@@ -28,6 +28,44 @@ export default class HttpRequestService {
...
+   *     // and will return false if not found
+   *     const isAccepted = this.request.accepts("text/xml"); // false
+   *
+   * @return boolean|string
+   *     False if the request doesn't accept any of the passed in types,
+   *     or the content type that was matches
+   */
+  public accepts (request: any, type: string|string[]): boolean|string {
+    const acceptHeader = request.headers.get('Accept')
+    const acceptedContentTypes = acceptHeader.split(';')[0];
```
Guergeiro, line 54: As I expressed in https://github.com/drashland/deno-drash/issues/173#issuecomment-613671510, a client might pass multiple things that the function could not handle...
```json
{
    "Accept": "application/json;text/html;"
}
```
```json
{
    "Accept": " application/json; text/html;"
}
```
```json
{
    "Accept": "application/json; text/html"
}
```
Now the results of these dummy jsons:
```typescript
// Example 1
console.log(acceptContentTypes); // Outputs ["application/json","text/html"]

// Example 2
console.log(acceptContentTypes); // Outputs [" application/json"," text/html"]

// Example 3
console.log(acceptContentTypes); // Outputs [" application/json"]
```
A working solution is to remove extra spaces with regex.
```typescript
const acceptedContentTypes = acceptHeader.replace(/\s/g, "").split(';')

// Example 1
console.log(acceptContentTypes); // Outputs ["application/json","text/html",""]

// Example 2
console.log(acceptContentTypes); // Outputs [" application/json","text/html",""]

// Example 3
console.log(acceptContentTypes); // Outputs [" application/json","text/html"]
```

Having the extra element in the array will not affect the rest of the implementation.

src/services/http_request_service.ts
```diff
@@ -28,6 +28,44 @@ export default class HttpRequestService {
...
+
+    // for when `type` is a string
+    if (typeof type === 'string')
+      return acceptedContentTypes.indexOf(type) >= 0 ? type : false;
+
+    // for when `type` is an array
+    const matches = type.filter(t => acceptedContentTypes.indexOf(t) >= 0);
```
Guergeiro, line 61: I would swap `.filter()` for `.find()`.
```typescript
const matches = type.find(t => acceptedContentTypes.indexOf(t) >= 0); // This contains the first element found.
return typeof matches != "undefined" ? matches : false;
```

example_app/request_accepts_resource_test.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Does not accept the type the resource expects
+  response = await members.fetch.get("http://localhost:1667/request-accepts", {
+    headers: {
+      "Accept": "text/html",
+    },
+  });
+  json = JSON.parse(await response.json());
+  await members.assert.equals(json.success, false);
+  await members.assert.equals(json.message, 'This resource only accepts JSON');
+});
```
Guergeiro, line 25: Maybe tests for preference `Accept` values with 3+ possible `Content-Type`, making sure it follows the client precedence?

src/services/http_request_service.ts
ebebbington: That implementation shouldn't be necessary right? I get your point actually if the Accept header is `application/json;text/html`, what I implemented would actually ignore the `text/html` part (didn't realise the content types can also be seperated by `;`). But because the `acceptedContentTypes` is just a string, even doing `' something, hello; text/html; application/json'.indexOf(type)` should still be able to capture if the `type` exists? Eg there shouldn't be a need to trim spaces or turn it into an array?

```
const acceptHeader = request.headers.get('Accept'); // "  text/html;0.5; someother stuff"
const split= acceptHeader.split(';'); // ["  text/html", "0.5", "someother stuff"];
const contentTypes = split[0]; // "  text/html"
contentTypes.indexOf("text/html") >= 0; // true
```

example_app/request_accepts_resource_test.ts
ebebbington: Good shout, working on reworking the tests to be more rigorous

src/services/http_request_service.ts
Guergeiro:
```typescript
const str = "hello; goodbye";
const array1 = str.split(";"); // ["hello", " goodbye"]
console.log(array1.indexOf("goodbye")); // Outputs -1 (didn't find element)
```
```typescript
const str = "hello; goodbye";
const array1 = str.replace(/\s/g, "").split(";"); // ["hello", "goodbye"]
console.log(array1.indexOf("goodbye")); // Outputs 1 (found element)
```
The reason I'm saying this is because, even though browsers remove all blank spaces in their requests, the client (for example, postman/postwoman) doesn't.

src/services/http_request_service.ts
ebebbington: I don't think i see much difference in using `.find()`, but I have incorporated part of your approach to make that block of code cleaner (yet to be committed) but looks like:
```
const matches = type.fnd(...); // the usual
return matches.length ? macthes[0] : false;
```
Cleaner code than what was before:
```
if (something)
  if (something)
     return something
return something else
```

src/services/http_request_service.ts
ebebbington: Yeah fair enough actually, if content types could be in 1st and 2nd index, I can't do `.split(';')[0]` so i'll have to just split, and remove the spaces then. Nice one!

Guergeiro: Oh, btw. In the *working solutions* I proposed, I blundered with copy and paste. There are no spaces in the example 2 and 3. Should be `["application/JSON","text/html"]` in both cases.

ebebbington: Had to change it up a bit as `str.replace(...).split(';')` might equal `["text/html,text/plain", "...", "..."]` and a `thatArray.indexOf("text/html")` wouldnt work

src/services/http_request_service.ts
Guergeiro: Your implementation is fine, honestly. Mine doesn't address any issue, just code style, since you are actually trying to find ONE element.

ebebbington:
Thanks @Guergeiro for all the useful input! :)

Issue #176, opened 2020-04-15
Support decorators for middleware in resources


crookse:
## Summary

What: Support decorators for middleware in resources.

Why: Currently, you define middleware in the resources using the resources' `middleware` property. It would be nice to define middleware using decorators and attaching those decorators to each HTTP method (if they require middleware).

## Acceptance Criteria

- [ ] Update documentation
- [x] Update tests
- [x] Decorators can be used in resources

## Example Pseudo Code (for implementation)

```typescript
import Drash from "https://deno.land/x/drash@v0.39.5/mod.ts";

class HomeResource extends Drash.Http.Resource {
  static paths = ["/"];

  @middleware("some middleware")
  public GET() {
    this.response.body = "Hello GET request!";
    return this.response;
  }

  @middleware("some other middleware")
  public POST() {
    this.response.body = "Hello POST request!";
    return this.response;
  }
}

const server = new Drash.Http.Server({
  address: "localhost:1447",
  response_output: "text/html",
  resources: [HomeResource]
});

server.run();
```

Guergeiro:
I like it. How will you do decorators?
```typescript
@middleware("some middleware")
@middleware("double middleware")
public GET() {
    this.response.body = "Hello GET request!";
    return this.response;
}
```
```typescript
@middleware(["some decorator","double middleware"])
public GET() {
    this.response.body = "Hello GET request!";
    return this.response;
}
```

crookse:
@Guergeiro,

Probably the second approach with the array

PR #181, opened 2020-04-16
Issue #176 decorators

ebebbington:
Supports decorators for one instance: middleware.

*Bare in mind, some of this is in it's early stages*

Will only support the following usage *currently*:
`@middleware('yourMiddleware')`

## Progress Notes
* A new decorators directory has been created to hold the method and logic for the `@middleware` method
* The logic inside the above is by all means not ready. I think it needs looking at to see how it could be implemented, suggestions welcome!
* The middleware method is now part of the exports in `mod.ts`
* `before_request` and `after_request` have been added to the `middleware` property for the server instantiation
* The server class has been updated to handle this new functionality, for before and after requests
* The middleware decorator file does need to be cleaned up as well, if it's possible to improve or move the logic within

## Problems I See
* On each before request and after request in the server class, those methods get all the middle-wares regardless if they are for *before* or *after*, I dont see how we could limit the logic to only get the before and after. 
* How we store the middlewares before server initialisation - middlewares are ready before they have access to the server, so currently i have kept them in a variable inside the `decorators/middle.ts` file. This isn't ideal.
* Maybe a specific look over from @crookse to check for any possible uncaught errors or security holes in the logic i added to thee server class
* Tests need to be written after some of these questions are figured out

## Example Usage from a Dev
```
import {Drash, Middleware } from "../mod.ts";

class TestResource extends Drash.Http.Resource {
    static paths = ["/test"];

    @Middleware(['AuthMiddleware', 'LogoutMiddleware'])
    //@ts-ignore - USER NEEDS A TSCONFIG.JSON
    public GET() {
        return this.response
    }
    @Middleware('AuthMiddleware')
    public POST() {
        ...
    }
}

class AuthMiddleware extends Drash.Http.Middleware {
    public run() {
        console.log('YOU ARE NOT AUTHED')
        //throw new Drash.Exceptions.HttpMiddlewareException(403, "NOT AUTHED");
    }
}
class LogoutMiddleware extends Drash.Http.Middleware {
    public run() {
        console.log('YOU ARE LOGGED OUT NOW')
    }
}

    address: 'localhost:1447',
    resources: [
        TestResource
    ],
    middleware: {
        resource_level: {
            before_request: [
                AuthMiddleware
            ],
            after_request: [
                LogoutMiddleware
            ]
        }
    }
});
server.run()
```


crookse:
I didn't notice it before, but I see now we have to define resource-level middleware in `before_request` and `after_request` blocks during server creation. That changes the syntax a bit. It does put it more in line with how the server middleware is configured though. Just putting this out there so we can all see it... if we kept the same syntax, the middleware would be defined like:

```
@Middleware('before_request:AuthMiddleware')
```

That's ugly.

ebebbington:
> I didn't notice it before, but I see now we have to define resource-level middleware in `before_request` and `after_request` blocks during server creation. That changes the syntax a bit. It does put it more in line with how the server middleware is configured though. Just putting this out there so we can all see it... if we kept the same syntax, the middleware would be defined like:
> 
> ```
> @Middleware('before_request:AuthMiddleware')
> ```
> 
> That's kind of weird IMO.

Yeah I agree with you. I did have to capitalise the decorator to avoid conflicts with the existing middleware class- don't know if this is a bad thing?

crookse:
I like it as `@Middleware`, so it's not a bad thing to me unless someone justifies why we should use `@middleware`

Guergeiro:
Honestly, if decorators become a problem, I would implement the Middleware using the Decorator Design Pattern and not TypeScript decorators.

ebebbington:
Closed in favour of https://github.com/drashland/deno-drash/pull/186

PR #186, opened 2020-04-18
Issue #176 decorators

Guergeiro:
My take on the whole TypeScript decorators. I couldn't get it to work with the existing `Drash.Http.Middleware` class... I went around the issue and used kinda the ExpressJS way.

```typescript
// Simple example

import { Drash } from "../deno-drash/mod.ts";
import { runMiddleware } from "../deno-drash/src/http/middleware.ts";

function LogMiddleware(req: any, server: Drash.Http.Server, response: Drash.Http.Response) {
    // User can access all this
    response.body = "123";
}

    static paths = ["/"];
    @runMiddleware({ before_request: [LogMiddleware] })
    public GET() {
        console.log("GET / Called");
        return this.response;
    }
}

    address: "localhost:1447",
    response_output: "text/html",
    resources: [HomeResource],
});
server.run();
```

Didn't do typings yet and added the changes to Drash.Http.Middleware for the lack of better option. This is all subject to change.


Guergeiro:
I summon @crookse and @ebebbington to test this 😄

src/http/middleware.ts
```diff
@@ -89,3 +89,34 @@ export abstract class Middleware {
    */
   abstract run(): any | void;
 }
+
+export type middlewareToRun = {
+  before_request?: any[];
+  after_request?: any[];
+};
```
ebebbington, line 96: Is it possible to do:
```
before_request?: Function[];
after_request?: Function[];
```
Just to be more explicit on the types?

src/http/middleware.ts
```diff
@@ -89,3 +89,34 @@ export abstract class Middleware {
...
+export function runMiddleware(obj: middlewareToRun) {
```
ebebbington, line 97: Wonder if it's possible to have a clean solution where this is named `middleware`? (Your approach might be able to do this compared to mine). @crookse might also need your input on this on how this might conflict with existing `middleware` methods/types etc and hopefully not cause any confusion

What about if we move server level middleware into decorators as well (outside of this feature maybe)? (think you mentioned it @crookse), that way it could reduce any conflicts or confusion with the `middleware` naming. And if possible, the `runMiddleware` method could check if the decorator is over a class or method

ebebbington:
I really like this approach, it's very well done!

My only gripe is if a dev wants a fair amount of middleware to run for a method, for example 4. I don't know how common this is but i think with having to define `before/after_request` it might end up looking like this:
```
@runMiddleware({before_request: [LogMiddleware, AuthMiddleware], after_request: [LogMiddleware, OtherMiddleware]})
public GET() {
    ...
}
```

src/http/middleware.ts
Guergeiro: Done. Even typed the functions for better typing support.

src/http/middleware.ts
Guergeiro: I agree the naming should be revised. I would prefer to do it after #185. I already started messing with the branch and it will improve our typings a lot! I'll keep a look onto this.

Guergeiro:
Some things to do.
- [ ] Remove *resource level middleware* implementations from [server.ts](https://github.com/Guergeiro/deno-drash/blob/middleware-as-decorators/src/http/server.ts)
- [ ] Remove *resource level middleware* tests from [middleware_tests.ts](https://github.com/drashland/deno-drash/blob/master/tests/unit/http/middleware_test.ts)
- [ ] Update *middleware* in [middleware_resource](https://github.com/drashland/deno-drash/blob/master/example_app/middleware_resource.ts)
- [ ] [Fix](https://github.com/drashland/deno-drash/blob/master/example_app/middleware_resource_test.ts)?

One of the things we gained from this implementation is we no longer need to check for Middleware declaration. This means that [http_middleware_exception.ts](https://github.com/drashland/deno-drash/blob/master/src/exceptions/http_middleware_exception.ts) is dead.

crookse:
If the test needs fixing then yea let's fix it. What I don't want to happen is we make the test pass the code instead of the code passing the test.

Guergeiro:
From https://github.com/drashland/deno-drash/pull/186#issuecomment-616007227
> If the test needs fixing then yea let's fix it. What I don't want to happen is we make the test pass the code instead of the code passing the test.

What I mean by removing is actually removing and doing new tests 😄

Since they do not work anymore, having it test will just fail. I will add them.

crookse:
Sounds good to me!

Guergeiro:
This is currently an example app. Tell me if you're okay with the implementation. Naming Types is still a bitch for me...

```typescript
import { Drash } from "../deno-drash/mod.ts";

async function BodyMiddleware(req: any, server: Drash.Http.Server, response: Drash.Http.Response) {
    const res = await fetch("https://api.github.com/users/Guergeiro/repos");
    const data = await res.json();
    response.body = JSON.stringify(data);
}
function LogMiddleware(req: any, server: Drash.Http.Server, response: Drash.Http.Response) {
    console.log("this is a log");
}

@Drash.Decorators.MiddlewareClass({ before_request: [LogMiddleware] })
    static paths = ["/"];
    @Drash.Decorators.MiddlewareMethod({ after_request: [BodyMiddleware] })
    public GET() {
        return this.response;
    }
    public PUT() {
        return this.response;
    }
}

    response_output: "application/json",
    resources: [HomeResource],
});

server.run({
    hostname: "localhost",
    port: 1447,
});

```

ebebbington:
Looks good to me, really really nice job on it! I don't suppose you could refrain from having your IDE auto format and changes made committed under a single commit? Just makes it really hard to see what's actually changed :/ Think we end up using `deno fmt` anyways.

Gonna leave the approval to @crookse

crookse:
Can we change `MiddlewareClass` to `ClassMiddleware` and `MiddlewareMethod` to `MethodMiddleware`? I feel like that's more semantic.

We can also go with `ResourceMiddleware` and `HttpMethodMiddleware`; unless being generalized with the decorator names is better? Thoughts?

Guergeiro:
From https://github.com/drashland/deno-drash/pull/186#issuecomment-616243214
> Looks good to me, really really nice job on it! I don't suppose you could refrain from having your IDE auto format and changes made committed under a single commit? Just makes it really hard to see what's actually changed :/ Think we end up using deno fmt anyways.

My bad on that one... I'll try to use it in the future. I'm not on my main workstations right now, but is it possible to add it in git hooks? That way, when people do `git add` it does it. Also, are you using `deno fmt` default config?

Guergeiro:
From https://github.com/drashland/deno-drash/pull/186#issuecomment-616246226

> Can we change `MiddlewareClass` to `ClassMiddleware` and `MiddlewareMethod` to `MethodMiddleware`? I feel like that's more semantic.

That's no problem. Like I said, naming typings (and others) is a bitch...

> We can also go with `ResourceMiddleware` and `HttpMethodMiddleware`; unless being generalized with the decorator names is better? Thoughts?

I don't find it a problem. But maybe for the sake of structure we should keep it under `Decorators` namespace? Honestly, I don't really know... I would like to make them just `Middleware` regardless of being at Class level or Method level. But I didn't know how, since when the decorator is invoked, it carries the "context" and that "context" is different if we're talking about Methods or Class.

> Can we choose one syntax over the other? In server-level middleware we use the `run()` method and extend `Drash.HttpMiddleware`. For resource-level middleware, we define small functions that take in `req`, `res`, and `server`. Can we just use functions instead and not have server-level middleware extend `Drash.Http.Middleware`? I like the idea of middleware just being functions that get injected into the application. That way devs can create maybe a `middleware.ts` file where they export all of their middleware; and then they use them where they want--at the server level or resource level.

I didn't want to mess with that server_level ones because, not only there were things that are using them, but also because I couldn't figure out how to do it in `public async handleHttpRequest(request: any): Promise<any>`.

> All in all it looks good! Very robust, flexible, and explicit. You want me to write the docs for this?

Don't yet. I believe we **should** do this the best way (core feature). **Must** be ready at least for `1.0.0` but not rushed.

crookse:
> Don't yet. I believe we **should** do this the best way (core feature). **Must** be ready at least for `1.0.0` but not rushed.

Ok. I'm just excited for this feature lol. Let's see if we can make the syntax for both resource-level and server-level the same. I'm going to checkout your PR and see if I can make that happen.

ebebbington:
> From [#186 (comment)](https://github.com/drashland/deno-drash/pull/186#issuecomment-616243214)
> 
> > Looks good to me, really really nice job on it! I don't suppose you could refrain from having your IDE auto format and changes made committed under a single commit? Just makes it really hard to see what's actually changed :/ Think we end up using deno fmt anyways.
> 
> My bad on that one... I'll try to use it in the future. I'm not on my main workstations right now, but is it possible to add it in git hooks? That way, when people do `git add` it does it. Also, are you using `deno fmt` default config?

I'm not sure actually, i don't think i've ever made a format commit for Drash yet (though there's times i should have done and have been corrected by Eric) but i think the standard in Deno is to format a file as a separate commit and name it `git commit -m "deno fmt"`

Didn't even realise `deno fmt` used a config lol so i guess i'll answer with no :P

Guergeiro:
New implementation...

```typescript
async function asyncFunc(req: any, response: Drash.Http.Response) {
    const res = await fetch("https://api.github.com/users/Guergeiro/repos");
    const data = await res.json();
    response.body = data;
}
function LogClass(req: any, response: Drash.Http.Response) {
    console.log("Class");
}
function LogMethod(req: any, response: Drash.Http.Response) {
    console.log("Method");
}
function ConvertToJson(req: any, response: Drash.Http.Response){
    response.body = JSON.stringify(response.body);
    response.header.set("Content-Type", "application/json");
}

@Drash.Decorators.Middleware({ before_request: [LogClass], after_request: [ConvertToJson] })
    static paths = ["/"];
    @Drash.Decorators.Middleware({ before_request: [LogMethod, asyncFunc], after_request: [LogMethod] })
    public GET() {
        return this.response;
    }
    public PUT() {
        return this.response;
    }
}

    resources: [HomeResource],
});

server.run({
    hostname: "localhost",
    port: 1447,
});
```

A limitation from TypeScript decorators is **class constructors** are always *"created"* before **methods**. This means the logs will be the following
- before request *Class->Method*
- after request *Class->Method*

Issue #187, opened 2020-04-18
Agnostic source code


crookse:
Just making a placeholder issue for right now. I'll fill this in when I have some time.

## Summary

What: The codebase should have no knowledge of what runtime is being used.

Why: This makes it easier to port to other JavaScript runtimes (e.g., Node).

## Acceptance Criteria

- [ ] All Deno.{someMember} come from `deps.ts`

## Example Pseudo Code (for implementation)

```typescript
import { RUNTIME } from "../deps.ts";

RUNTIME.readFileSync("file.txt")
RUNTIME.readAll("file.txt")
RUNTIME.open("file.txt");
RUNTIME.Reader
RUNTIME.SomeOtherThing
```

PR #190, opened 2020-04-19
Issue #187 Agnostic Source Code

crookse:
* export RUNTIME from deps.ts
* change test ports so they don't conflict


crookse:
closing. we're not supporting other runtimes.

PR #195, opened 2020-04-21
decorators

crookse:
Example app:

```typescript
import { Drash } from "../../deno-drash/mod.ts";

class ResourceLevelMiddleware extends Drash.Http.Middleware {
  public async run() {
    // Before the request
    if (!this.response) {
      console.log("Executing resource-level middleware before the request");
      return;
    }
    if (this.response) {
      console.log("Executing resource-level middleware after the request");
    }
  }
}

class ServerLevelMiddleware extends Drash.Http.Middleware {
  public run() {
    // Before the request
    if (!this.response) {
      console.log("Executing server-level middleware before the request");
      return;
    }
    if (this.response) {
      console.log("Executing server-level middleware after the request");
    }
  }
}

  static paths = ["/"];

  @Drash.Decorators.HttpMethodMiddleware(["ResourceLevelMiddleware"])
  public GET() {
    this.response.body = "GET request received!";
    return this.response;
  }

  public POST() {
    this.response.body = "POST request received!";
    return this.response;
  }
}

    response_output: "application/json",
    resources: [HomeResource],
    middleware: {
      server_level: [
        ServerLevelMiddleware
      ],
      resource_level: [
        ResourceLevelMiddleware
      ]
    }
});

server.run({
    hostname: "localhost",
    port: 1447,
});


console.log("Server started at localhost:1447");

```

### DIFF
--- a/mod.ts
+++ b/mod.ts
@@ -6,6 +6,9 @@
 // Compilers
 import { TemplateEngine as BaseTemplateEngine } from "./src/compilers/template_engine.ts";
 
+// Decorators
+import { HttpMethodMiddleware as http_method_middleware } from "./src/decorators/http_method_middleware.ts";
+
 // Dictionaries
 import * as log_levels from "./src/dictionaries/log_levels.ts";
 import mime_db from "./src/dictionaries/mime_db.json";
@@ -49,6 +52,10 @@ export namespace Drash {
 
   export namespace Compilers {
     export class TemplateEngine extends BaseTemplateEngine { }
+  }
+
+  export namespace Decorators {
+    export const HttpMethodMiddleware = http_method_middleware;
   }
 
   export namespace Dictionaries {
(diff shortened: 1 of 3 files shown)


Guergeiro:
[Both](https://github.com/drashland/deno-drash/pull/186) approaches are good for me. Now it comes down to syntax. I prefer the fact that the developer knows that a function will be run before/after request without having to worry about coding the logic.

From https://github.com/drashland/deno-drash/pull/195#issue-406431468
> ```ts
> const server = new Drash.Http.Server({
>     response_output: "application/json",
>     resources: [HomeResource],
>     middleware: {
>       server_level: [
>         ServerLevelMiddleware
>       ],
>       resource_level: [
>         ResourceLevelMiddleware
>       ]
>     }
> });
> ```

This is what I really dislike though. Having to declare *middleware* at server start is really counter-intuitive from a design pattern point of view... For me, we should only declare *middleware* at server start if that *middleware* is directly related to the server.

All in all, and I'm okay if you prefer pushing the *middleware* feature like that :smile:

crookse:
this PR sacrifices A LOT of performance. closing.

PR #201, opened 2020-04-22
Added version property to Drash namespace

ebebbington:
Added a `version` property to the `Drash` namespace.

Also updated the `console/typescript/bump_versions` script to include the updating of the string in `mod.ts`

Example usage:

```typescript
import { Drash } from "https://deno.land/x/drash@v0.41.0/mod.ts";
console.log(Drash.version) // "v0.41.0"
```

### DIFF
--- a/mod.ts
+++ b/mod.ts
@@ -63,7 +63,7 @@ export namespace Drash {
    *
    * @property string version
    */
-  export const version: string = "v0.41.0";
+  export const version: string = "v0.41.1";
 
   export namespace Decorators {
     export type MiddlewareFunction = MiddlewareFunctionDefinition;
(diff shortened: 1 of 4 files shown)


ebebbington:
Currently writing the tests

mod.ts
```diff
@@ -56,6 +56,15 @@ export namespace Drash {
...
   }
 
+  /**
+   * @description
+   *     Drash version. Also represents what Deno version is
+   *     supported.
+   *
+   * @property string version
+   */
+  export const version: string = "v0.41.0";
```
crookse, line 66: we're currently at v0.41.1

crookse: ok i get what you're saying now.. so this is a problem. the bump version script will need to handle this file differently because although we match deno's version, we can still push out fixes and increment the patch version.

ebebbington: Changed it to a manual change, which isn't a problem because it adds about 10s onto the update process

PR #216, opened 2020-05-22
upgrade deno to v1.0.1

ebebbington:
Upgrades deno to v1.0.1 from v1.0.0 (std 0.50.0)

**Description**

* Updated all references to deno's and drash's version to reflect the new one

**Other Notes**

* Any other useful information goes here

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -46,7 +46,7 @@ Run the following command.
 console/bump_verisons
 ```
 
-Update the version property in `mod.ts`:
+Update the version property in `mod.ts` to Deno's (as shown in their releases):
 ```typescript
 export const version: string = "the new version here"; // "v0.41.1"
 ```
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@
 - [Documentation](#documentation)
 - [Features](#features)
 - [Why use Drash?](#why-use-drash)
-- [Example Drash App](#example-drash-app)
 - [Contributing](#contributing)
 - [License](#license)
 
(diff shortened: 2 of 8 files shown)


ebebbington:
tests have started to fail:
```
Download https://deno.land/std@v0.52.0/path/_constants.ts
error: EOF while parsing a value at line 1 column 0
```
now i'm not getting the above and getting the following for all resource tests (coffee resource example):
```
failures:

CoffeeResource
Server listening: http://localhost:1667
Http: error sending request for url (http://localhost:1667/coffee): error trying to connect: tcp connect error: Connection refused (os error 61)
    at unwrapResponse ($deno$/ops/dispatch_json.ts:43:11)
    at Object.sendAsync ($deno$/ops/dispatch_json.ts:98:10)
    at async fetch ($deno$/web/fetch.ts:591:27)
    at async file:///.../deno-drash/example_app/coffee_resource_test.ts:5:14
    at async asyncOpSanitizer ($deno$/testing.ts:36:5)
    at async Object.resourceSanitizer [as fn] ($deno$/testing.ts:70:5)
    at async TestApi.[Symbol.asyncIterator] ($deno$/testing.ts:264:11)
    at async Object.runTests ($deno$/testing.ts:346:20)
```

ebebbington:
Above 3 commits fixed the issue(s) (wierdly), and removed the example app from the README in upgrade-deno as it wasn't needed anymore

ebebbington:
@crookse I've found that, say i am on deno v1.0.0 and upgrade the v1.0.1, right after this if i run tests, we get the error that the build is showing (`error sending request to url`), but this happens _sometimes_...

Deno v1.0.1 also has a bug where error messages send to stderr do not show the actual message, and instead it shows: `error: `. This is happening on the last test (`tests/test.ts`)

If I upgrade to deno v1.0.2, the tests seem to be completely fine - so i suggest we maybe upgrade to deno v1.0.2 as well? Which I think makes sense as i believe most users would upgrade to v1.0.2 to bypass the bug in v1.0.1

Issue #182, opened 2020-04-16
create-app Script


ebebbington:
## Summary

***Inspiration came from [Dinja](https://github.com/denjucks/dinja)***

What: Create a `create-app` script that allows a user to create a basic file structure from the CLI, that would also write content into those files. Can include a structure for an API or Web app

Why: It'll be a neat feature for a user to do. Similar to how React and Angular do as well, to quickly create a basic structure for a Drash app.

## Additional Information

Here is some information that can help implement this feature. Open to changes or to give your own twist on it. Mainly to ensure we supply a script that a user can use to customise their project :

* Script can be executed from the command line like:
```
deno --allow-read [--as-api|--as-app] https://deno.land/x/drash/create_app.ts
```
* A `--as` flag can be passed in, to specify if the user wants an API-like structure or a web app. The script would determine what type of contents to write to the files and what files to write based on this
```
deno ... --as-api ...
deno ... --as-app ...
```
* A `--with` flag can be passed in for usage with vue. This will determine to write basic vue files
```
deno ... --as-app --with-vue ...
```
* Script should send message to stderr if `--with-vue` is provided without `--as-app`
* Script should send message to stderr if no valid (api, app) `--as` flag is supplied
* Script name is `create_app.ts` (open for discussion on changes)
* Script could lie in the root directory of Drash
* The script could create the following files regardless of flags:
    * app.ts
    * deps.ts
    * resources/
         * example_resource.ts
    * config.ts
    * tests/
        * example_test.ts
* Whilst writing the files, produce the output to stdout so a user knows something is happening
* Maybe have a boilerplate directory that can be used to copy?

## Acceptance Criteria (Guidelines)

- [ ] A create-app script is created to build a boilerplate for drash
- [ ] Documentation updated to include this
    
## Example Pseudo Code (for implementation)

```typescript

```

PR #217, opened 2020-05-24
Issue #182 create app script

ebebbington:
Fixes [#182](https://github.com/drashland/deno-drash/issues/182)

**Description**

* Fulfilled the requirements inside of the issue

* Added a boilerplate directory instead of creating a gazillion strings 

* Added test

* Supports usage of api, web app and web app with vue

**Other Notes**

* Would the README need to be updated to say 'here you can do _this_, just run: `deno .... create_app.ts`

* Decided to ignore the feature of echoing out each file that was created - there doesn't seem a need for it and the script finishes so quickly

* To run it yourself, just checkout the branch and:
`cd ../; mkdir test-proj; cd test-proj; deno run --allow-read --allow-run --allow-write ../deno-drash/create_app.ts [--api] [--web-app] [--with-vue]`


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
-        "experimentalDecorators": true
+        "experimentalDecorators": true,
+        "module": "ESNext"
     }
 }
```
Guergeiro, line 6: I there a reason we are targeting ESNext? Because remember, when you do ESNext, we might be targeting features that TypeScript already compiles to, but Deno doesn't understand.

create_app.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+//////////////////////////////////////////////////////////////////////////////
+// FILE MARKER - ENTRY POINT LOGIC ///////////////////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+// Requirement: Now allowed to ask for an API AND Web App
+if (wantsApi && wantsWebApp) {
+    Deno.run({
+        cmd: ['echo', red('--web-app and --api options are now allowed to be used together. Use the --help option on how to run this script')]
```
crookse, line 137: Use the --help option to see how to run this script.

create_app.ts
```diff
@@ -0,0 +1,179 @@
...
+    })
+    Deno.exit(1)
+}
+
+// Requirement: One main argument is required
+const tooFewArgs = !wantsHelp && !wantsWebApp && !wantsApi
+if (tooFewArgs) {
+    Deno.run({
+        cmd: ['echo', red('Too few options were given. Use the --help option on how to run this script')]
```
crookse, line 146: See comment above

create_app.ts
```diff
@@ -0,0 +1,179 @@
...
+    })
+    Deno.exit(1)
+}
+
+// Requirement: --with-vue is only allowed to be used with --web-app. Helps for user error mainly
+if (wantsVue && !wantsWebApp) {
+    Deno.run({
+        cmd: ['echo', red('The --with-vue option is only allowed for use with a web app. Use the --help option on how to run this script')]
```
crookse, line 154: See comment above

create_app.ts
```diff
@@ -0,0 +1,179 @@
...
+    Deno.run({
+        cmd: ['echo', green(message)]
+    })
+}
+
+/**
+ * Send our thank you message for using it
+ */
+function sendThankYouMessage () {
+    notesForUser.push('Run your application: deno run --allow-run --allow-net --allow-read app.ts')
```
crookse, line 80: do we need --allow-run here?

console/create_app/resources/home_resource.ts
```diff
@@ -0,0 +1,25 @@
+import { Drash } from "../deps.ts";
+
+export default class HomeResource extends Drash.Http.Resource {
+    static paths = ["/"];
+
+    public GET() {
+        this.response.body = this.response.render('/index.html');
```
crookse, line 7: do you think we should make another `home_resource.ts` file for the `--api` option? Otherwise, this will respond with `false` when we go to `localhost:1667`

ebebbington: Yeah, it makes sense, added here: https://github.com/drashland/deno-drash/pull/217/commits/4d04ccfc3c9a8247eb5a7f3d43717a76fa9693f1

create_app.ts
ebebbington: No, not sure why that is in there then but have removed it: https://github.com/drashland/deno-drash/pull/217/commits/4d04ccfc3c9a8247eb5a7f3d43717a76fa9693f1

PR #231, opened 2020-06-08
Improved ci

### LINKED ISSUE #145: Improve version bumping workflow
## Summary

What: Improve the workflow of bumping version numbers in the codebase.

### LINKED ISSUE #228: CI To Check for New Deno Versions
Placeholder issue until more information is gathered around this new feature

## Summary
As quoted by @Guergeiro:

> My vision would be to have a CI that checks for a Deno version every day. When a new version is found, it creates a new branch and opens a PR. Then we accept it or fix problems if it exists. That way we don't import it, it just stays in the CI.

What: Explain ...

Why: Explain ...

## Acceptance Criteria

- Add acceptance criteria as bulleted list

```typescript
// Add example pseudo code for implementation
```

Guergeiro:
Fixes #145
Fixes #228 

**Description**

- Added ci to auto bump Deno versions and create a Pull Request.
- It will run everyday at 00:00 UTC.
- It will only open a PR if it changes the `deps.ts` file.

**Other Notes**
- Also changed the behaviour of `master.yml` to separate testing and linting.

- It is using `CI_USER_PAT`. I don't know if deno-drash secret PAT is @crookse or @drashbot. It should be @drashbot, but I can change the variable name easily.


crookse:
The CI user is drashbot :)

Guergeiro:
@crookse @ebebbington everything working. We should see a PR today at 00:00 UTC because I'm using a version of [dmm](https://github.com/drashland/dmm) that flags our `deps.ts` as upgradable. We update **dmm** in the future.

prisis:
Why you not using the https://github.com/denolib/setup-deno action?

And can you test it on linux, windows and macos?  
  
You can take a look on https://github.com/dmpjs/deno-module-template/blob/master/.github/workflows/continuous-integration.yml   
  
And if you need help with workflows, just ping me

crookse:
I didn’t know about setup deno action. 😮

Guergeiro:
Thanks for your input @prisis. I've heard of denolib action template. Honestly it's so easy to configure Deno that it's almost not required to have a action template for that 😉

We will keep Deno matrix to a single version since Deno std is still not stable, so a bunch of tests fails every time we upgrade Deno. Will definetly add more versions as soon as Deno launches v1.0.0 of std library.

As far as OS goes, it's not an issue to have multiple OS testing (denolib will be useful for this since deno_install is different from OS to OS).

Guergeiro:
@prisis updated to use denolib/setup-deno. Honestly, better just prepare the CI to be scalable than to have to updated again when the time comes.

Guergeiro:
@crookse I ignore `create_app` tests since we have coded them using unix style path and not windows style... You know `/home/user` vs `C:\User\`. The slashes are the problem.

PR #236, opened 2020-06-09
Update dependencies

crookse:


ebebbington:
Closing because this shows the bumper ci works, but we need to address a problem dmm has, which removed the `v` (as seen in the changes made above). I will release a new version with the fix now and update the time for the cronjob to run

PR #250, opened 2020-06-12
Issue #232 improve create app tests

### LINKED ISSUE #232: Assert stdout and stderr for Create App Tests
## Summary

What: Expand upon the current create app tests to not only stop bloating the shell that the tests are ran in, but to assert the stdout and stderr response for running this script, for each test. 

Why: Because this script is CLI based, there is a lot of information given, be it stdout or stderr. It would improve the testing if we can assert the correct responses are given for the different types of scenarios

## Acceptance Criteria

* For each test, assert that the stdout and stderr is correct

```typescript
// The usual stuff here, but add `stdout` and `stderr` to piped
const p = Deno.run({
    ...
    stdout: "piped",
    stderr: "piped"
})
// Usual tests that already exist
...
// Response tests
const output = await p.output() // Uint8Array
const errors = await p.stderrOutput() // Uint8Array
const outputString = new TextDecoder("utf-8").decode(outputstring)
const errorsString = ... // same as above
members.assertEquals(outputString, ...)

// Should we get errors for bad resource id or leaking resources, you can do:
//. await p.close()
//. const output = await p.output()
// Or remove the `await.close()` entirely as `p.output()` closes anyways (but it can be flakey)
```

You may see things like:
```
- = actual | + = expected
- "Something..."
+ "
+ "Something..."
```
If you do, it might because the stdout/stderr has a space at the end, or it needs a new line, so for example:
```
assertEquals(outputString, 
    'Doing something...\n' + 
    'Something else... \n'
)
```

ebebbington:
Fixes #232 

**Description**

* Added assertions for stdout and stderr

* Removed one test because it isn't actually needed and covered by the test that requires a main argument (a main argument is either: --help, --api or --web-app and not --with-vue)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -36,7 +36,9 @@ jobs:
           deno test --allow-all --config tsconfig.json tests/integration/app_3004_pretty_links/tests.ts
 
       - name: Create APP
-        run: deno test tests/cli/create_app_test.ts --allow-read --allow-write --allow-run
+        run: |
+          deno cache mod.ts
+          deno test tests/cli/create_app_test.ts --allow-read --allow-write --allow-run
 
   linter:
     strategy:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Create APP
         run: |
-          deno cache mod.ts
+          deno cache create_app.ts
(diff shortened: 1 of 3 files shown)


ebebbington:
Need to cache the script because the first test is recieving "Compiling .../.create_app.ts" as stdout

PR #265, opened 2020-06-24
Use Rhum

crookse:


tests/test_deps.ts
```diff
@@ -0,0 +1 @@
+export { Rhum } from "https://deno.land/x/rhum@v1.0.0/mod.ts";
```
Guergeiro, line 1: This is something I wanted, keep the tests deps separate from Drash! Other than that, I'm fine with using Rhum as our testing framework.

🚀

ebebbington: Thats a good idea, i like it too

PR #273, opened 2020-06-28
Issue #271 respect accept header

### LINKED ISSUE #271: Accept header values not being considered when generating a response
**Describe the bug**

**To Reproduce**

1. create the following application

```typescript
// File: app.ts

import { Drash } from "https://deno.land/x/drash/mod.ts";

  static paths = ["/"];
  public GET() {
    this.response.body = "Hello";
    return this.response;
  }
}

  logger: new Drash.CoreLoggers.ConsoleLogger({
    enabled: true,
    level: "debug",
  }),
  resources: [HomeResource],
});

server.run({
  hostname: "localhost",
  port: 1447,
});

console.log("Server listening: http://localhost:1447");

```

2. run the app
3. Go to localhost:1447 in insomnia or postman
4. Send a GET request (expected response content-type header should be application/json)
5. Send a GET request with the Accept header of text/html (expected content-type header should be text/html, but it's not)

**Expected behavior**

The Accept header should be considered when generating a response. The response object sets its content-type header in it's `constructor()`. Currently, it still uses the `response_content_type` code that was removed in an earlier version. This code needs to be removed and replaced with the code that parses the Accept header.

crookse:
Fixes #271 

This PR does the following:

* moves accepts() from HttpRequestService to HttpService
* removes logic around `response_content_type` and respects the `accept` header
* removes tests for `response_content_type`


src/http/response.ts
```diff
@@ -74,15 +82,46 @@ export class Response {
...
     this.headers = new Headers();
     this.template_engine = options.template_engine;
     this.views_path = options.views_path;
-    this.headers.set("Content-Type", request.response_content_type);
+    this.headers.set("Content-Type", this.getContentType());
   }
 
   // FILE MARKER: METHODS - PUBLIC /////////////////////////////////////////////
 
+  protected getContentType(): string {
```
ebebbington, line 95: We able to have a doc block for this?

ebebbington:
I like what you've done though, how you've written the logic for the getContentType method

src/http/response.ts
crookse: done!

crookse:
> Get the content type from the request objects "Accept" header

mad the change. yea i should've been more verbose here. also changed the method name

crookse:
> Maybe add: `If no default response content type was passed in, then defaults to application/json`, and also something like `Returns the first accepted content type if multiple as within the accept header` somewhere?

made the change

PR #284, opened 2020-07-01
[uniform-docblocks] Uniform docblocks

ebebbington:
Address any doc blocks that are not uniformed, and do not follow Deno's standards


ebebbington:
I guess I may as well covert every doc block to tsdoc then if i'm changing one part of the code base?

PR #285, opened 2020-07-01
chore: Update release.yml

tbaumer22:
**Description**

Without upgrading the CLI to the latest version, there could be errors when publishing for future releases.

* Added `eggs upgrade` to upgrade the CLI after installing

**Other Notes**

Thanks so much for adding Drash to nest.land!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,5 +15,6 @@ jobs:
      - run: |
           cd eggs
           export PATH="/home/runner/.deno/bin:$PATH"
+          eggs upgrade
           eggs link --key ${{ secrets.CI_NESTLAND_API_KEY }}
           eggs publish
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,6 @@ jobs:
     steps:
      - uses: actions/checkout@v2
      - uses: denolib/setup-deno@master
-       with:
-         deno-version: v1.x
-
      - run: deno install -A -f --unstable -n eggs https://x.nest.land/eggs@0.1.5/mod.ts
      - run: |
           cd eggs
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
      - uses: actions/checkout@v2
      - uses: denolib/setup-deno@master
        with:
-         deno-version: 1.1.0
+         deno-version: v1.x
 
      - run: deno install -A -f --unstable -n eggs https://x.nest.land/eggs@0.1.5/mod.ts
      - run: |


Guergeiro:
Why is this required? As far as I know, even if we upgrade the cli, we are still dependent on the deno version, so we can't blindly upgrade it...

tbaumer22:
This is required because we could introduce breaking changes to the way our module uploading system works in the future. I suppose you could install the latest Deno version for this as well to be sure that nothing will break. The CLI doesn't need to be able to compile the module for it to be uploaded to the registry, so upgrading Deno wouldn't pose any issues (that I see) for this workflow.

Guergeiro:
So if I always target the latest version of Deno, will the Eggs version work as well?

Imagine (all versions in this example are hypothetical) you released a new version that is compatible with Deno 1.5. Good, I'm running the last version of Deno and upgrade to last version of Eggs. Next day, Deno releases 1.6 that breaks Eggs. Now the workflow fails because the cli doesn't install on that version of Deno.
Or in another situation, I upgrade the Eggs version but the one that I upgraded is not compatible with the Deno version I have installed. Boom, right?

tbaumer22:
These are very good points. We actually have CI running each night to ensure that our CLI will be compatible with a nightly Deno version. In the event that the CI fails, we prepare for the new version and fix the new bugs. This means that it is very likely that the latest CLI version will be compatible with the latest stable Deno version, especially considering the fact that Drash won't be publishing a new version every couple hours.

You can see our CLI nightly testing [here](https://github.com/nestlandofficial/nest.land/blob/master/.github/workflows/eggs-test-deno-nightly.yml)

Guergeiro:
Okay. I trust that. Is there a way to target eggs latest version/master? If so, something like this would be nice:
```yaml
name: release
on:
  release:
    types: [published]
jobs:
  publish-egg:
    runs-on: ubuntu-latest
    steps:
     - uses: actions/checkout@v2
     # Using latest deno
     - uses: denolib/setup-deno@master

     # Target eggs latest/master?
     - run: deno install -A -f --unstable -n eggs https://x.nest.land/eggs@latest/mod.ts
     - run: |
          cd eggs
          export PATH="/home/runner/.deno/bin:$PATH"
          eggs link --key ${{ secrets.CI_NESTLAND_API_KEY }}
          eggs publish
```

tbaumer22:
Unfortunately we don't support that, as we purposefully don't let any modules use any sort of implicit versioning. We find that it tends to break things over time. While our CLI would be an exception to this, there is a possibility that we could get backlash for allowing something internally that is prohibited for anyone else.

Although, I'm not completely against considering this for Eggs, as I think it would simplify things for the CLI. For now, however, installing Eggs `0.1.5` and running `eggs upgrade` will get the same result.

Sorry for the confusion!

tbaumer22:
Okay, it should be good to merge now. Thanks for the patience!

PR #291, opened 2020-07-05
Update dependencies

drashbot:

### DIFF
--- a/.github/workflows/bumper.yml
+++ b/.github/workflows/bumper.yml
@@ -20,6 +20,11 @@ jobs:
       - name: Update Dependencies
         run: deno run --allow-net --allow-read --allow-write https://deno.land/x/dmm/mod.ts update
 
+      - name: Update Test Dependencies
+        run: |
+          cd tests
+          deno run --allow-net --allow-read --allow-write https://deno.land/x/dmm/mod.ts update
+
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v2
         with:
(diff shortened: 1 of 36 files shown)


ebebbington:
Test deps needed to be updated too, and I have added a section into `DEV.md` to include notes around upgrading deno

ebebbington:
Unfortunately dmm won’t update the test deps because it’ll only update a file named “deps.ts” which I think the test deps file is named “test_deps.ts” - think dmm could add support for this but we can discuss it @crookse @Guergeiro

Guergeiro:
That's why I changed from test_deps.ts to deps.ts.

ebebbington:
Ah, it removed the commits from this PR as well...I don't remember who originally worked on this (hard to tell from the above) but I'll try get it sorted before tomorrows bump (unless @Guergeiro gets to it first i'm not fussed, just want to get it out of the way :P )

Guergeiro:
It's fine. I still think we should make dmm pass a flag that allows the user to choose the name of his `deps.ts` file. We might use the convention, but someone might not. Something like `dmm update --file=myWeirdNameDepsFile.ts`

ebebbington:
> It's fine. I still think we should make dmm pass a flag that allows the user to choose the name of his `deps.ts` file. We might use the convention, but someone might not. Something like `dmm update --file=myWeirdNameDepsFile.ts` Of course this should be cross commands, that is, we can use on check and info.

FYI for PR documentation purposes as I mentioned this in the discord, i created the issue here: https://github.com/drashland/dmm/issues/36

PR #292, opened 2020-07-05
Jsdocs to tsdocs

### LINKED ISSUE #289: Convert JSDoc to TSDoc
## Summary

What: Convert JSDoc blocks (though we use a mix of JSDoc and Deno's standard), to TSDoc blocks in Typescript files

Why: We primarily use TypeScript so it makes sense to use TSDoc. Types are already documented in the method/property anyways.

## Acceptance Criteria

- [ ] All doc blocks converts to TSDoc

```
* @param string filename
*        The filename to do something with

* @param filename - The filename to do something with
```

## Resources

TSDoc is very similar, there is really only a few minor changes to make to each doc block

https://github.com/microsoft/tsdoc

Guergeiro:
Fixes #289 

**Description**
- Updated src to TSDocs


src/core_loggers/logger.ts
```diff
@@ -70,12 +58,13 @@ export abstract class Logger {
...
   /**
-   * @description
-   *     Write a log message. All extended classes must implement this method.
-   *     See Drash.CoreLoggers.ConsoleLogger/FileLogger for examples.
+   * Write a log message. All extended classes must implement this method.
+   * See Drash.CoreLoggers.ConsoleLogger/FileLogger for examples.
    *
-   * @param any logMethodLevelDefinition
-   * @param string message
+   * @param logMethodLevelDefinition {@link ./interfaces/log_level_structures.ts#LogLevelStructure | Interface of Log}
```
crookse, line 64: should we change all of the `logMethodLevelDefinition` descriptions to use this description?

src/core_loggers/logger.ts
```diff
@@ -210,18 +186,15 @@ export abstract class Logger {
...
   /**
-   * @description
-   *     Send the message to the write method (which should be in the child
-   *     class).  Also, do some prechecks before sending to see if the log
-   *     message should be written.
+   * Send the message to the write method (which should be in the child
+   * class).  Also, do some prechecks before sending to see if the log
+   * message should be written.
+   *
+   * @param logMethodLevelDefinition {@link ./interfaces/log_level_structures.ts#LogLevelStructure | The dictionary definition of the log message's level.}
```
crookse, line 193: see https://github.com/drashland/deno-drash/pull/292/files#r450133125

src/http/request.ts
```diff
@@ -66,22 +66,20 @@ export class Request extends ServerRequest {
   //////////////////////////////////////////////////////////////////////////////
 
   /**
-   * @see Drash.Services.Http.Service.accepts()
+   * Used to check which headers are accepted
+   * @param type It is either a string or an array of strings that contains the Accept Headers
```
crookse, line 70: missing `-` in between param name and description

src/http/resource.ts
```diff
@@ -1,69 +1,47 @@
...
+   * A property to hold the middleware this resource uses.
    *
-   *     All derived middleware classes MUST define this property as static
-   *     (e.g., static middleware = ["MiddlewareClass"];)
-   *
-   * @property string[] middleware
+   * All derived middleware classes MUST define this property as static
+   * (e.g., static middleware = ["MiddlewareClass"];)
    */
   public middleware: { after_request?: []; before_request?: [] } = {};
```
crookse, line 14: making an issue to remove this logic completely. we're not using a static property on resources anymore to define resource-level middleware.

src/http/resource.ts
```diff
@@ -1,69 +1,47 @@
...
-   *     (e.g., static paths = ["path"];)
+   * A property to hold the paths to access this resource.
    *
-   * @property string[] paths
+   * All derived resource classes MUST define this property as static
+   * (e.g., static paths = ["path"];)
    */
   public paths: string[] = [];
 
   public paths_parsed: Drash.Interfaces.ResourcePaths[] = [];
```
crookse, line 31: woops. i forgot to add this docblock when i did the linting stuff

```
A property to hold the expanded versions of the paths. An expanded version of a path looks like the following:
    {
      og_path: "/:id",
      regex_path: "^([^/]+)/?$",
      params: ["id"],
    }
```

crookse:
btw this PR will break our api_reference.json file that's used by our documentation pages. i think we should stop parsing our own API reference data and just use doc.deno.land. we can add make the API Reference links in our docs point to doc.deno.land. for example, in the docs, API Reference > Http > Server would link to https://doc.deno.land/https/deno.land/x/drash/src/http/server.ts

src/http/resource.ts
Guergeiro: I already removed them from the docs. Will handle the issue next.

src/core_loggers/logger.ts
Guergeiro: I just followed the TSDocs convention. Though I have no way to tell if it's the best practice...

Guergeiro:
> btw this PR will break our api_reference.json file that's used by our documentation pages. i think we should stop parsing our own API reference data and just use doc.deno.land. we can make the API Reference links in our docs point to doc.deno.land. for example, in the docs, API Reference > Http > Server would link to https://doc.deno.land/https/deno.land/x/drash/src/http/server.ts

Hum... Damn... And the fact the deno doc uses JSDoc and not TSDoc... Is it worth the change?

src/http/resource.ts
Guergeiro: Where do you want that? On `paths` or `paths_parsed`?

crookse:
> Hum... Damn... And the fact the deno doc uses JSDoc and not TSDoc... Is it worth the change?

I think we should do it. If doc.deno.land doesn't parse our doc blocks correctly, then we can revert the change. won't know for sure what's going to happen unless we merge this branch.

src/http/resource.ts
crookse: Paths parsed

src/core_loggers/logger.ts
crookse: Haha alright. It’s fine then. I’m resolving this

PR #310, opened 2020-07-11
Issue #277 remove any in interface

### LINKED ISSUE #277: linting - remove any in ParsedRequestBody interface
What: Improve `ParsedRequestBody` interface by removing `any` type. See screenshot below for context about this issue.

Screen Shot 2020-06-28 at 11 36 43

Why: Improve the quality of the codebase.

Acceptance Criteria:

If this issue is possible to overcome:

- [ ] remove `any`
- [ ] tests pass

ebebbington:
Fixes #277 

**Description**

* Improved the interface for parsedBodyRequest to match it's actual expected data

* Due to this, a lot of types in the request class had to be changed, but as a bonus, there were no lint errors after making these changes

### DIFF
--- a/deps.ts
+++ b/deps.ts
@@ -36,6 +36,7 @@ export {
 export {
   FormFile,
   MultipartReader,
+  MultipartFormData
 } from "https://deno.land/std@v0.60.0/mime/multipart.ts";
 
 export {
--- a/deps.ts
+++ b/deps.ts
@@ -36,7 +36,7 @@ export {
 export {
   FormFile,
   MultipartReader,
-  MultipartFormData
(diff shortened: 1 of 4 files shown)


ebebbington:
Also, i removed the `let ret: ... = { content_type: "", data: undefined }` from `parseBody` - this was causing some wierd errors with the different types of data returned. Anyways, i moved this kind of logic into each `if (content type is this)` because we can better type the return data

ebebbington:
Note that this did break some tests for getting body files, but it wouldn't affect users (eg how users use drash, whereas the tests are calling an 'internal' method)

PR #319, opened 2020-07-15
use v1.x branch in import statement

crookse:


crookse:
It works. I tested it with Rhum.

PR #342, opened 2020-07-24
Clean up

crookse:
**Description**

* Removes unused files
* Cleans up deps
* Updates Drash version in egg.json
* Fixes create_app to build from local cache instead of fetching deno-drash@master and then building (b48c60d148df65a340f00937ae0d8931ef36bf5c, 51dd4821fbc81a5e3bb029e48720a8b305d751b0, 33345c4c47d3d2710f32fec122889b931575b478)


crookse:
I think the tests fail because create_app is targeting master, and master doesn't have the updated deps file (the one in this PR). So it's testing using a new deps file (the one in this PR) against the old deps file (the one in master). I think PRs that update the create_app script's deps file will always fail.

tests/deps.ts
```diff
@@ -1,5 +1,9 @@
 export { Rhum } from "https://deno.land/x/rhum@v1.0.1/mod.ts";
 export * as path from "https://deno.land/std@v0.61.0/path/mod.ts";
+export {
+  assertEquals,
+  assertThrows,
+} from "https://deno.land/std@v0.61.0/testing/asserts.ts";
```
ebebbington, line 6: Isn't this covered by `Rhum.asserts`?

console/create_app/deps.ts
```diff
@@ -1,2 +1,2 @@
-export { Drash } from "https://deno.land/x/drash@v1.0.0/mod.ts";
-export { assertEquals } from "https://deno.land/std@v0.52.0/testing/asserts.ts";
+export { Drash } from "https://deno.land/x/drash@v1.1.1/mod.ts";
+export { assertEquals } from "https://deno.land/std@v0.61.0/testing/asserts.ts";
```
ebebbington, line 2: The first test that failed is because we changed this, but the tests haven't been reflected, eg tests still look for v1.0.0 drash, and the other tests fail because when a test fails, it doesn't clean up any files/dirs made, which is why any test after says "... exists"

tests/deps.ts
crookse: oh. yeah.

crookse: 3e17b2d

ebebbington:
Oh actually, the stuff you've updated for all the versions, could you add those to our workflow (if they arent already added)?

console/create_app/deps.ts
crookse: so what's a solution so we don't run into this issue when we update this file?

crookse:
> Oh actually, the stuff you've updated for all the versions, could you add those to our workflow (if they arent already added)?

yea. good one

console/create_app/deps.ts
ebebbington: I guess there isn't much we could do, unless adding a `cleanUpEnv()` method or something (i think it already exists) at the top of **each test**, so when a test does fail, the next one(s) would still run

crookse:
> Oh actually, the stuff you've updated for all the versions, could you add those to our workflow (if they arent already added)?

Updated:

https://github.com/drashland/castle/blob/master/creating_a_github_release.md#publishing-releases

console/create_app/deps.ts
crookse: alright, since this is a clean up PR, i may as well put that work into this PR

crookse:
> Fix the tests. Tests will always run in the current code, that is, if you add a new workflow, it will be run in the pull request.

The tests weren't going to pass because of this:

https://github.com/drashland/deno-drash/blob/543e3d9c3415ab3b13d5a237957586e722d7ad0c/create_app.ts#L24

It would always check against what's in master and not what's in the current code. @ebebbington's original implementation of using `import.meta.url` was on the right track. @ebebbington, I'm using your original implementation to fix the tests. Using `importa.meta.url` seems to be the more robust way of handling this script and how it copies all of the boilerplate stuff.

crookse:
tests are now fixed: b48c60d

console/create_app/deps.ts
crookse: b48c60d, 51dd482, 33345c4

PR #361, opened 2020-08-13
[fix-default-content-type] Fix response content type for browsers

ebebbington:
**Description**

* There was an edge case for browsers and their accept header. In that edge case, even if the content type config was set in the server, the response content type would still try match the requests Accept header. Take this resource:
```
class Some resource {
  public GET () {
    this.response.body = {}
    return this.response
}
```
Fetch requests to that resource would return {}, whereas browser requests would return null, and the response content type of a browser request would be set to `text/html` - eg it matched the 'first type' on getting content type from accept header method.

Now, i created a browser request example test and adjusted the response class to fix this, so it better follows the. flow of:

1. Use content type in config 
2. else if not set, get type from accept header
3. if resource methods set content header explicitly, don't override that
3. else default to application/json


ebebbington:
Think i might add some extra docs on explaining how it calculates the content type in more detail (is this pr is approved)

src/http/response.ts
```diff
@@ -90,10 +90,18 @@ export class Response {
     this.headers = new Headers();
     this.template_engine = options.template_engine;
     this.views_path = options.views_path;
-    this.headers.set(
-      "Content-Type",
-      this.getContentTypeFromRequestAcceptHeader(),
-    );
+    if (this.options && this.options.default_response_content_type) {
+      //const defaultContentType = this.options.default_response_content_type ?? "application/json";
```
prisis, line 94: is this needed?

ebebbington: The comment? No i think i left that inn by mistake :X Remove it now though

PR #365, opened 2020-08-30
Update dependencies

drashbot:

### DIFF
--- a/.github/workflows/bumper.yml
+++ b/.github/workflows/bumper.yml
@@ -30,6 +30,9 @@ jobs:
           cd console/create_app
           deno run --allow-net --allow-read --allow-write https://deno.land/x/dmm/mod.ts update
 
+      - name: Update Deno Version Occurances
+        run: deno run  --allow-net --allow-net --allow-write console/update_deno_version_strings.ts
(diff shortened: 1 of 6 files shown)


ebebbington:
@crookse @Guergeiro i'm proposing we use a script to automate any other manual tasks we need to do - thoughts?

Guergeiro (dismissed):
I agree with it. I think the bot removed the script though.

ebebbington:
Yeah it did lol, got it all local anyways so will just push it up again

@crookse what do you think about using a script for the repos, to update any other manual work when bumping? eg update deno version in workflows, std version refs in dmm's README

crookse:
@ebebbington sounds good to me

ebebbington:
ok good for a final review

Guergeiro (approved):
This seems good (I'm on phone), but essentially we still need to manually run the script right?

ebebbington:
It's ran by the bumper CI, so the ci will do it's usual (run dmm update), then run that script

Should note that if it runs and there isn't a newer version released, nothing happens (ie no errors, no code change)

PR #368, opened 2020-09-16
Add pr ci

ebebbington:
**Description**

* Adds a CI that automatically updates/creates a draft release for each merged PR

* Each repo has it's own 'variation' of release content, so actually with the templates for the release drafter, it makes it easier -  for example dmm's release content is slightly different

* On making a PR, make sure:
    * The title is in the format of "feat: <title>" or "fix: <title>". This is because the. PR title is what's shown in the release, and to support the feat/fix/etc prefix to changes, we do that here
    * Use the related labels. If pr is a patch (fix), use patch label, if a minor (feat) then use minor label, etc. If the PR isn't worth being mentioned in the release, then just leave the labels blanks, for example on readme updates

* When the PR is merged, the draft we get created if it isn't already, and will display all PR's merged since the last release

**Other Notes**

* It just mimics our current process, but instead:
    * it's automated
    * we must follow a format for pr titles and labels

* I haven't found a way yet we can dynamically use the Deno version in the template (see config i believe). It'd be good to do something like:
```
template: |
    * Uses Deno version v${latestDenoVersion}
```
```

### DIFF
--- a/.github/release_drafter_config.yml
+++ b/.github/release_drafter_config.yml
@@ -14,7 +14,15 @@ tag-template: 'v$RESOLVED_VERSION'
 #       - 'bug'
 #   - title: '🧰 Maintenance'
 #     label: 'chore'
+
 change-template: '- $TITLE (#$NUMBER)'
+
+# Only add to the draft release when a PR has one of these labels
+include-labels:
+  - 'major'
+  - 'minor'
+  - 'patch'
+  - 'chore'
 
 # Here is how we determine what version the release would be, by using labels. Eg when "minor" is used, the drafter knows to bump up to a new minor version
 version-resolver:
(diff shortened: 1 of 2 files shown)


ebebbington:
This is also more os a proposal

Guergeiro:
This is really good! I'm just wondering how we could avoid having it run on `master` every time, because that branch is like a placeholder for all stuff... Maybe we could have another branch called `release`? How would this work? When we feel like there are enough features/fixes/etc we make a PR to this branch.

ebebbington:
We actually don't have to do anything, if a PR is merged with either `minor|major|match|chore` labels, the draft release gets updated - it won't actually change/modify anything

There's an example you can use and see in the copyright-free repo

Guergeiro (approved):
Oh yeah, this just drafts a new release, doesn't actually publish it!

ebebbington:
:) And we can still review the draft before releasing too to make sure everything is ok, just helps remove one manual step we have to do

Issue #378, opened 2020-09-28
Unable to Get Objects From `getBodyParam`


ebebbington:
**Describe the bug**
TSC errors whenn trying to get an object from `getBodyParam`

**To Reproduce**
```
const param = this.request.getBodyParam("user")
if (param) {
  const user = { ...param }
}
```

**Expected behavior**
No TSC errors

**Suggested Solution(s)**
`getBodyParam` tells us it returns either a string or undefined, but using the realworld app as an example, we can get objects from a body params, which isn't supported in the `getBodyParam` method.

Maybe we could add this possibility to the return type

**Screenshots**
![image](https://user-images.githubusercontent.com/47337480/94386989-c20b7e80-0140-11eb-96b2-5b06603f83fe.png)

![image](https://user-images.githubusercontent.com/47337480/94387001-c768c900-0140-11eb-8b92-b322cb89b490.png)

PR #385, opened 2020-10-05
fix: Dont return null if body param is false; Support body params being data types other than a string

### LINKED ISSUE #384: `getBodyParam` returns `null` when a param is set to `false
**Describe the bug**
Set a body param to `false`
Retrieving it will yield `null`

**Expected behavior**
It should be false

ebebbington:
Fixes #384 and #378 

**Description**

* Passing in a body param that is `false` will not return `null`, and instead return it's expected value
* Assume that a body param can be of any type and not just a string

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -119,7 +119,7 @@ export class Request extends ServerRequest {
    *
    * @returns The corresponding parameter or null if not found
    */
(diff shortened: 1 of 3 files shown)


ebebbington:
closing because http create app tests dont like hashes in branch name, opened up another: #386

PR #393, opened 2020-10-13
chore: make resource lookup faster

crookse:
## Description

* Makes lookup for resources faster without using a loop over an array of resources. Resources now use the `IndexService` class in https://github.com/drashland/services/tree/master/index repo.
* Caches request URL paths and associates them with resources. This makes subsequent requests to the same URLs faster.
* Adds benchmarks in the CI.
* Organizes `addHttpResource()` -- puts parsed paths logic into `protected` methods under semantic method names

## Benchmarks

Command used: `wrk -c 40 -d 10 http://localhost:1447/10000` (to search for the last resource in the `Map`)
App type: Hello world
Number of resources loaded into the server: 10k

```
MASTER BRANCH
526 requests in 10.07s, 50.85KB read
Requests/sec:     52.21

THIS BRANCH
284449 requests in 10.10s, 26.86MB read
Requests/sec:  28161.97
```

Command used: `wrk -c 40 -d 10 http://localhost:1447`
App type: Hello world
Number of resources loaded into the server: 1

```
MASTER BRANCH
281967 requests in 10.00s, 22.86MB read
Requests/sec:  28191.94

THIS BRANCH
285477 requests in 10.01s, 23.14MB read
Requests/sec:  28511.35
```


ebebbington:
@crookse dont suppose you could add what the results were before, just so we can compare them again? Sorry

crookse:
> @crookse dont suppose you could add what the results were before, just so we can compare them again? Sorry

done. it's in the description

ebebbington:
> > @crookse dont suppose you could add what the results were before, just so we can compare them again? Sorry
> 
> done. it's in the description

Cheers, sorry to be a pain, just thought it’d be easier to understand

src/http/server.ts
```diff
@@ -53,6 +53,12 @@ export class Server {
    */
   public port: number = 1447;
 
+  /**
+   * A property to hold all paths associated with their resources for lookups
+   * during the request-resource lifecycle.
+   */
+  protected paths: any = {};
```
Guergeiro, line 60: Can't we strongly type this? If not, please change it to `unknown`. Let's start avoiding `any`.

src/http/server.ts
```diff
@@ -758,32 +758,35 @@ export class Server {
...
-        if (pathMatchesRequestPathname == null) {
-          continue;
-        }
-
-        pathMatchesRequestPathname.shift();
-        const pathParamsInKvpForm: { [key: string]: string } = {};
-        pathObj.params.forEach((paramName: string, index: number) => {
-          pathParamsInKvpForm[paramName] = pathMatchesRequestPathname[index];
-        });
+    let resource = undefined;
```
Guergeiro, line 761: We should avoid using `let`, if possible, because it leads to a possible variable assignment deep in the logic that we might forget about. This could be avoided with early return statements.
If possible, I would prefer if a quick refactor in this function.

src/http/server.ts
```diff
@@ -758,32 +758,35 @@ export class Server {
...
+      const pathParamsInKvpForm: { [key: string]: string } = {};
+      resource.paths_parsed.forEach(
+        (pathObj: Drash.Interfaces.ResourcePaths) => {
+          pathObj.params.forEach((paramName: string, index: number) => {
+            pathParamsInKvpForm[paramName] = pathMatchesRequestPathname[index];
+          });
+        },
+      );
+      request.path_params = pathParamsInKvpForm;
+      break;
```
Guergeiro, line 786: This break statement is here because...? Basically we only iterate one time the for loop.

Guergeiro (changes requested):
Some nitpicking by my part. I would say priority is handling the `break` statement, followed by strongly typing `paths` and at last (and this is really nitpicking) the function refactor (which I'm okay with since it's more about style).

src/http/server.ts
```diff
@@ -758,32 +758,35 @@ export class Server {
...
+    for (const regex in this.paths) {
+      if (regex === "/" && request.url_path === "/") {
+        resource = this.paths[regex];
+        break;
+      }
 
-        request.path_params = pathParamsInKvpForm;
-        return resource;
+      const pathMatchesRequestPathname = request.url_path.match(
+        regex,
```
saragee3, line 769: nitpick: do we have a linter rules for this? why is this singular arg on a new line? imo doesn't make it more readable. same with line 779, seems like we should have a bit of consistency here esp because 780 has 2 args on the same line?

src/http/server.ts
saragee3: this makes sense. do we want to move toward a functional approach in general? 

@crookse can you link me to the test file for this function?

src/http/server.ts
crookse: 43bb3a4

src/http/server.ts
crookse: @Guergeiro, 00c7219

@saragee3, we could go with a functional approach, but this is the only place this kind of iteration happens? here's the test file and the test suite that covers this method: https://github.com/drashland/deno-drash/blob/7be08137fa581c7089f604117aa06ee48d1af585/tests/unit/http/server_test.ts#L55

src/http/server.ts
crookse: we iterate until we find a matching resource. if the resource is found, then we return that resource. if you look further up the method, you'll see see this:

```typescript
      const pathMatchesRequestPathname = request.url_path.match(
        regex,
      );
      if (pathMatchesRequestPathname == null) {
        continue;
      }
```

if the current regex path is not matched with the requested path, then we know that the resource associated with the regex path is not the resource, so we continue and iterate again

src/http/server.ts
crookse: the formatting is done by the `deno fmt` command. that command is also in the CI process

crookse: i should also mention that the CI fails if your code doesn't pass the `deno fmt` command

src/http/server.ts
Guergeiro: Hummm... I see.

crookse: do you think a while loop would be faster? lol...

crookse: @Guergeiro @ebebbington @saragee3 , i just tried the code using a while loop. not seeing much of a difference. if anything it's a little bit slower. maybe while loop is faster using a larger array?

ebebbington:
@crookse remember to. add a `major`/`minor`/`patch`/`chore` label :) and prefix pr title with `<type>: ...`

src/http/server.ts
Guergeiro: That might be just some weird stuff. In terms of speed, they're both O(n).

saragee3 (dismissed):

src/http/server.ts
crookse: now using a string as an index. we search the string for indices that are associated with resources. once an index is found, we can use that index to `.get()` on the resource Map.

crookse:
putting this on hold. going to make an index service in drashland/services repo

crookse:
note to self: count slashes

```
{
   'test///': TestResource
}
```

crookse:
this is ready for review. i'm not adding anymore changes

src/http/server.ts
```diff
@@ -758,31 +760,246 @@ export class Server {
...
+    if (split.length > 1) {
+      if (urlWithoutParam[0] == "" || urlWithoutParam[0] == "/") {
+        urlWithoutParam = "/" + split[1];
+      } else {
+        urlWithoutParam = split[0];
       }
     }
+
+    try {
+      return this.resource_index_service!.getItem("\\^" + urlWithoutParam);
```
crookse, line 999: this should return an object like the following

```
{
  id: 1,
  search_term: some regex path,
  search_input: some path,
  result: the value of the id in the lookup table
```

this prevents us from having to first find the index of an item and then taking that index and searching the lookup table. the index service should do all this for us. if we search for something, then it should return results from the lookup table and information about the search (the search input used and the id of the item in the lookup table, etc)

src/http/server.ts
```diff
@@ -758,31 +760,246 @@ export class Server {
...
+      request.url_path,
+    );
+
+    // No resource found? GTFO.
+    if (!resourceLookupInfo) {
+      return resource;
+    }
+
+    if (resourceLookupInfo) {
+      resourceLookupInfo.forEach((result: ISearchResult) => {
```
crookse, line 789: we should use a while loop here for performance

src/http/server.ts
crookse: 5a694d2

src/http/server.ts
crookse: 5a694d2

src/exceptions/invalid_path_exception.ts
```diff
@@ -0,0 +1,20 @@
...
+  //////////////////////////////////////////////////////////////////////////////
+
+  /**
+   * Construct an object of this class.
+   *
+   * @param message - (optional) The exception message.
+   */
+  constructor(message?: string) {
+    if (!message) {
+      message = "Path must to be a string.";
```
ebebbington, line 16: oh just came across this - typo?

src/http/server.ts
```diff
@@ -556,114 +576,52 @@ export class Server {
...
+        paths = this.getResourcePaths(path);
       }
-    }
 
-    resourceClass.paths_parsed = newPaths;
+      resourceParsedPaths.push(paths);
+
+      // Include the regex path in the index, so we can search for the regex
+      // path during runtime in `.getResourceClass()`
+      this.resource_index_service.addItem(paths.regex_path, resourceClass);
```
crookse, line 621: adding an item should take in an array of items so that search terms can be built in one call

src/http/server.ts
```diff
@@ -758,32 +752,241 @@ export class Server {
...
+  /**
+   * Has the request URL path in question been handled previously? That is, a
+   * resource was found for it and an association was stored in the CACHED
+   * resource lookup Map.
+   *
+   * @param urlPath - The request URL path in question.
+   *
+   * @returns True if handled previously; false if not.
+   */
+  protected requestUrlWasHandledPreviously(urlPath: string): boolean {
```
crookse, line 943: no need for this (just put this one liner in where it's being called)

src/http/server.ts
```diff
@@ -758,32 +752,241 @@ export class Server {
...
+    return this.cached_resource_lookup_table.has(urlPath);
```
crookse, line 944: index service should handle caching

src/http/server.ts
```diff
@@ -758,32 +752,241 @@ export class Server {
...
+   *
+   * @param urlPath - The request URL path to match to a resource.
+   * @param index - The index to search. This index contains associations
+   * between all regex paths that the server is handling and indices to
+   * resources in the resource lookup Map.
+   *
+   * @returns Lookup information on a resource or undefined if not found in the
+   * index that is used for the lookup information.
+   */
+  protected getResourceLookupInfo(
```
crookse, line 959: this doesn't return lookup info. it returns matching resources. this method name should be changed to reflect that.

src/http/server.ts
```diff
@@ -758,32 +752,241 @@ export class Server {
...
+    urlPath: string,
+  ): Map<number, ISearchResult> {
+    const url = urlPath.split("/");
+
+    if (url[url.length - 1] == "") {
```
crookse, line 964: theses array manipulations should be documented with comments so that developers know what's going on

src/exceptions/invalid_path_exception.ts
crookse: 3cdf23caf75817

src/http/server.ts
crookse: finding: it was found that drash having its own cache was faster than the index service managing a cache

src/http/server.ts
crookse: disregard last comment. we're now using the IndexService. see the description in this PR for more information about how that service works. it's documented

src/http/server.ts
crookse: removed in 0203af5; no more caching logic needed (index service handles caching)

src/http/server.ts
crookse: removed in 0203af5 (no need for this since code was consolidated)

src/http/server.ts
crookse: these have been documented. see 0203af5. also, see getResourceClass()

src/http/server.ts
crookse: @Guergeiro, still reviewing?

src/exceptions/invalid_path_exception.ts
```diff
@@ -0,0 +1,20 @@
...
+   * Construct an object of this class.
+   *
+   * @param message - (optional) The exception message.
+   */
+  constructor(message?: string) {
+    if (!message) {
+      message = "Path must be a string.";
+    }
+    super(message);
+  }
```
Guergeiro, line 19: Syntax sugar:
```ts
constructor(message: string = "Path must be a string."){
  super(message);
}
```

Guergeiro (approved):
Looks good to me.

src/exceptions/invalid_path_exception.ts
crookse: 7c74004

Issue #382, opened 2020-10-03
Add .getAll*() method on request object


besingamkb:
## Summary
https://drash.land/drash/#/tutorials/requests/handling-url-query-params
What: Nice to have a `.all()` methods or similar on the `this.request` for handling all query params.

Why: easier to handle a lot of  parameters so you can assign it on a variable..
```ts
const params = this.request.all();
params['id'];
params['name'];
params['email'];
...
```

## Acceptance Criteria
Edit by @ebebbington below:

- [x] (#396) A `getAllUrlQueryParams()` method on `Drash.Http.Request` that returns `{[key: string]: string|null} | null`
    - If no params passed in, return null, 
- [x] (#404) `getAllBodyParams()`, `getAllHeaderParams()`, `getAllPathParams()`
- [x] Tests
- [x] Documentation (https://github.com/drashland/website/issues/211) (to be handled by Drash Land maintainers)

ebebbington:
I agree with this, nice one @besingamkb, i've just edited the issue description to include tests for acceptance criteria

MVEMCJSUNPE:
Can I work on the tests?

ebebbington:
Hey @MVEMCJSUNPE, to work on the tests you'll need to implement the feature - feel free to take this issue on :)

MVEMCJSUNPE:
@ebebbington I'll take this issue on then; looking at getUrlQueryParam, I don't think I'll have too much trouble with writing getAllUrlQueryParams.

ebebbington:
@besingamkb Can we have some confirmation on what you liked to see? Because the title and description feels very confusion, what would you like to see:
```typescript
const all = this.request.all()
console.log(all) // { headers: ..., queryParams: ..., etc etc }
// or
const queryParams = this.request.getAllQueryParams()
const queryStrings = this.request.getAllQueryStrings()
// etc etc
```

besingamkb:
@ebebbington i think that will work as long as we get all the data from any kind of request verb. the main point of this feature request is to make it easier for devs to access data. so instead of doing it one by one with `this.request.getBodyParam('test1')` we can do it this way.

```ts
const queryParams = this.request.getAllQueryParams();
queryParams['test1']
queryParams['test2']
// etc etc
```

crookse:
would the following suffice?

```typescript
// Get EVERYTHING
request.all() // => { query_parms: { ... }, header_params: { ... }, body_params: { ... } }

// Only get query params
[...]

PR #396, opened 2020-10-22
feat: Add getAllUrlQueryParams method

MVEMCJSUNPE:
Fixes [#382]

**Description**

getAllUrlQueryParams takes as input url query params and returns a record whose keys are the query params and whose values are the corresponding values of the query params. If no params are given, then null is returned. 

**Other Notes**
The jsdoc documentation for getAllUrlQueryParams probably could be worded a little better.

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -175,7 +175,16 @@ export class Request extends ServerRequest {
   }
 
   public getAllUrlQueryParams(...inputs: string[]): {[key: string]: string|null} | null {
-    return null;
+    if (inputs.length === 0) {
+      return null;
+    }
+    
+    let result: Record<string, string> = {};
+    for (const input of inputs) {
+      result[input] = this.url_query_params[input];
+    }
+
+    return result;
   }
 
   /**
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -174,6 +174,10 @@ export class Request extends ServerRequest {
     return null;
   }
 
+  public getAllUrlQueryParams(...inputs: string[]): {[key: string]: string|null} | null {
+    return null;
+  }
+
   /**
    * Get this request's URL path.
    *
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -178,8 +178,7 @@ export class Request extends ServerRequest {
    * Gets a record whose keys are the request's url query params specified by inputs
    * and whose values are the corresponding values of the query params.
    * 
-   * @param inputs url query params
-   * @returns null if no inputs or record containing key-value pairs
+   * @returns Key value pairs of the query param and its value. Empty object if no query params
    */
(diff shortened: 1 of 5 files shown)


ebebbington:
Noticing a problem here, http tests fail for forks because the branch doesn't exist on drashland/deno-drash

MVEMCJSUNPE:
Yeah, sorry about that. I tried pushing a commit unrelated to 'getAllUrlQueryParams', thinking I could fix it (which I probably shouldn't have done), so feel free to ignore it.

ebebbington:
That's my fault to be honest, i should have mentioned it

Actually, Drash already has a way to get all query params, i think we could just wrap around it.

Maybe you could make `Drash.Http.Request#getUrlQueryParams` private, then make a new method:
```
public getAllUrlQueryParams (): {[key: string]: string {
   return this.url_query_params
}
```

Then in the request tests,  replace `getUrlQueryParams` with `getAllUrlQueryParams`

Feel free to comment out the running of the create app tests in `.github/workflows/master.yml` for this PR (and commit it)

MVEMCJSUNPE:
Maybe it's just me, but it seems like it would fail all of the tests in the tests section of the Acceptance Criteria in #382 ? Should I just delete the tests I wrote, since you want me to replace getUrlQueryParams with getAllUrlQueryParams in the request tests?

ebebbington:
The original issue has gotten confusing, i think the focus is that this PR will add a method to get all url query params, by by all, i mean: 
```
GET /users --> 0 params
GET /users?name="ed" --> one param
GET /users?name="ed"&city="uk" --> 2 params
```
Most of the tests already exist as we have unit tests for the `getUrlQueryParams` method

And i guess only remove tests you feel are uneeded

src/http/request.ts
```diff
@@ -174,6 +174,17 @@ export class Request extends ServerRequest {
     return null;
   }
 
+  /**
+   * Gets a record whose keys are the request's url query params specified by inputs
+   * and whose values are the corresponding values of the query params.
+   * 
+   * @param inputs url query params
+   * @returns null if no inputs or record containing key-value pairs
```
ebebbington, line 182: Just need to remove the `@param` line and adjust the `@returns` description to say `Key value pairs of the query param and it's value. Empty object if no query params`,  just to be inline with TSDoc

ebebbington (changes requested):
Just a minor comment

src/http/request.ts
MVEMCJSUNPE: This is fine.

MVEMCJSUNPE (commented):
LGTM

ebebbington:
Thanks @MVEMCJSUNPE, nice work! :) It's greatly appreciated

ebebbington (approved):

PR #397, opened 2020-10-23
Use bumper service

crookse:
**Description**

* Uses `BumperService`
* Adds tests that test the `BumperService` is working as expected in this module
* Large `deno fmt` change (did `deno fmt` change the way it formats? looks like it's putting things in alphabetical order)


ebebbington (approved):
LGTM

PR #401, opened 2020-10-25
fix: Content length was 2 when content type is JSON and body is empty

ebebbington:
**Description**

* Found as part of getting cors middleware to work. Content length on response was 2 even when body was empty. This is because (see code changes), when `this.body` is empty, and you try `JSON.stringify` it, the result will be:
```
console.log(JSON.stringify("")) // """"
```
It essentially adds data and thus a body. So the body is **actually** empty, but when it comes to encoding that, it's a Uint8Array with a length of four, because we are pretty much doing `new TextEncoder().encode('""')`, which is wrong

image

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -196,7 +196,7 @@ export class Response {
 
     switch (contentType) {
       case "application/json":
-        return JSON.stringify(this.body);
+        return this.body ? JSON.stringify(this.body) : "";
       case "application/xml":
       case "text/html":
       case "text/xml":


crookse (approved):

ebebbington:
@crookse this might need to be reflected in the services repo?

crookse:
yea. i haven't stripped out anything yet. so this stuff will be def be in when drash migrates. also, i need to make that migration issue

ebebbington:
Alright, lemme know if you need me to do anything whilst you're out, cause you don't have access to an amazing UX on a mac :P

Issue #403, opened 2020-10-26
Address Create App Http Tests Failing For Forks


ebebbington:
## Summary

What: Http create app tests fail for for forks. Discuss best way to approach them

MVEMCJSUNPE:
Can't we just use the environment variable "GITHUB_REPOSITORY", which returns the owner and repo in the form of a string `${owner}/{${repo}`? This way, the programmer making the pull request will have the tests run on his/her fork on his/her own repo.

ebebbington:
That might work, feel free to give it a shot :)

PR #407, opened 2020-10-27
bugfix: Made a mistake when trying to fix deno url

MVEMCJSUNPE:
The official repo name was drashland/deno-drash,
but for some reason, I wrote drashland/deno-land.
My bad.

Fixes [#403 ]

**Description**

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -41,10 +41,10 @@ jobs:
           deno cache create_app.ts
           deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env
 
-      # Commented out since this doesn't work on forks of main repository
-      # - name: Create App (http)
-      #   run: |
-      #     deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
+      Commented out since this doesn't work on forks of main repository
+      - name: Create App (http)
+        run: |
+          deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
 
   linter:
     # Only one OS is required since fmt is cross platform
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -41,7 +41,6 @@ jobs:
           deno cache create_app.ts
           deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env
 
-      Commented out since this doesn't work on forks of main repository
       - name: Create App (http)
         run: |
           deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
--- a/tests/cli/create_app_test_http.ts
+++ b/tests/cli/create_app_test_http.ts
@@ -19,11 +19,11 @@ if (!latestBranch) {
 }
 
 if (!githubRepo) {
-  githubRepo = "drashland/deno-land";
+  githubRepo = "drashland/deno-drash";
 }
 
-const drashUrl = "https://raw.githubusercontent.com/" + githubRepo +
-  `/${latestBranch}`;
+const drashUrl =
+  `https://raw.githubusercontent.com/${githubRepo}/${latestBranch}`;
 
 function getOsCwd() {
   let cwd = `//${originalCWD}/console/create_app`;
@@ -176,6 +176,7 @@ Rhum.testPlan("create_app_test.ts", () => {
       async () => {
         const testCaseTmpDirName = tmpDirName + (tmpDirNameCount += 1);
         // Create new tmp directory and create project files
+        console.log(`${testCaseTmpDirName}`);
         Deno.mkdirSync(testCaseTmpDirName);
         const p = Deno.run({
           cmd: [
--- a/tests/cli/create_app_test_http.ts
+++ b/tests/cli/create_app_test_http.ts
@@ -14,12 +14,9 @@ const decoder = new TextDecoder("utf-8");
 let latestBranch = Deno.env.get("GITHUB_HEAD_REF");
 let githubRepo = Deno.env.get("GITHUB_REPOSITORY");
 
-if (!latestBranch) {
-  latestBranch = "master";
-}
-
-if (!githubRepo) {
+if (!githubRepo || !latestBranch) {
   githubRepo = "drashland/deno-drash";
+  latestBranch = "master";
 }
 
 const drashUrl =
--- a/tests/cli/create_app_test_http.ts
+++ b/tests/cli/create_app_test_http.ts
@@ -14,12 +14,20 @@ const decoder = new TextDecoder("utf-8");
 const githubRepo = Deno.env.get("GITHUB_REPOSITORY") ?? "drashland/deno-drash";
 let latestBranch = Deno.env.get("GITHUB_HEAD_REF") ?? "master";
 
+console.log("START");
+console.log(latestBranch);
+console.log(githubRepo);
+console.log(Deno.env.get("GITHUB_BASE_REF"));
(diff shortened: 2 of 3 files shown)


tests/cli/create_app_test_http.ts
```diff
@@ -176,6 +176,7 @@ Rhum.testPlan("create_app_test.ts", () => {
       async () => {
         const testCaseTmpDirName = tmpDirName + (tmpDirNameCount += 1);
         // Create new tmp directory and create project files
+        console.log(`${testCaseTmpDirName}`);
```
ebebbington, line 179: Just need to remove the debugging line :)

ebebbington (changes requested):
It was my bad, it got past code review

ebebbington:
@MVEMCJSUNPE you will need to uncomment the block from `.github/workflows/master.yml` to ensure the create app http tests work too

MVEMCJSUNPE:
Didn't notice those. Thanks! Is there a way to run tests in releases.yml locally?

ebebbington:
> Didn't notice those. Thanks! Is there a way to run tests in releases.yml locally?

https://github.com/nektos/act BUT it doesn't work with deno, i've tried and tested it and raised an issue, might work with simpler workflow files

.github/workflows/master.yml
```diff
@@ -41,10 +41,10 @@ jobs:
           deno cache create_app.ts
           deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env
 
-      # Commented out since this doesn't work on forks of main repository
-      # - name: Create App (http)
-      #   run: |
-      #     deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
+      Commented out since this doesn't work on forks of main repository
```
ebebbington, line 44: Just need to remove this line ;)

ebebbington: unresolving

ebebbington: No you have to make the code change yourself and commit it, the reviewing on GH is just a way to highlight code in a PR and comment in it

ebebbington:
@MVEMCJSUNPE the http tests seem to still be pulling through the drashland/deno-drash repo (and thus failing):

image

ebebbington:
Seems the `githubRepo` is not defined, or the value it's assigned is the drashland repo, you could try logging out the details in the test, which i. funily enough asked to remove earlier, i'm guessing you left it in to debug, but i forgot :P

MVEMCJSUNPE:
With the latest commit, I think we can be sure that githubRepo is simply being assigned to the drashland repo.

crookse:
> With the latest commit, I think we can be sure that githubRepo is simply being assigned to the drashland repo.

so just for future reference, the only modules allowed in deno-drash, rhum, dmm, dawn, and sockets are deno std modules or drashland modules. i see you've added octokit. is there a reason why a `fetch()` call can't be used to see if a URL to a branch returns a 404 status code?

MVEMCJSUNPE:
I guess I was just lazy. When I tried "https://raw.githubusercontent.com/drashland/deno-drash/master", it would give me a 404 error. While could try "https://raw.githubusercontent.com/drashland/deno-drash/master/create_app.ts" that seems kind of iffy to me: if a filename gets changed or remove, how would we know the difference between whether the repository not existing or the file not existing on the repository due to major changes?

crookse:
that's why we have acceptance criteria and tests to cover the codebase and cases like this in the CI. if we change the create_app script name, then it will break the integration tests in the CI. there's currently work being done to move the create app script into a drash CLI, and if it's not handled properly in all places in the codebase then the CI will fail

MVEMCJSUNPE:
For some reason, it's working on my computer:
create_app_success

tests/cli/create_app_test_http.ts
```diff
@@ -7,33 +7,20 @@
...
-}).then((branches: any) => {
-  if (
-    !branches.data?.find((branch: any) => {
-      branch.name === latestBranch;
-    })
-  ) {
-    latestBranch = "master";
-  }
+fetch(drashUrl + "/create_app.ts").catch((error) => {
+  drashUrl = "https://raw.githubusercontent.com/drashland/deno-drash/master";
```
ebebbington: Race condition here probably, do:
```
try {
  await fetch(drashUrl + "/create_app.ts")
} catch (err) {
  drashUrl = ...
}
```

ebebbington:
But it's probably failing due to the env vars inside the GH runner

ebebbington:
Oh ok cause this PR is under drashland, the GH runner env vars are pointing to this repo

ebebbington:
@MVEMCJSUNPE can you add:
```
console.log("START")
console.log(latestBranch)
console.log(githubRepo)
console.log("END")
```
Below line 15ish in `tests/cli/create_app_http_test.ts` and push it up?

ebebbington:
Maybe we could do something like `const url = 'GITHUB OWNER + "deno-drash" + GITHUB_BRANCH'`

MVEMCJSUNPE:
By "GITHUB OWNER" and "GITHUB_BRANCH", are you talking about environment variables? I checked Github's [environment variables](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables), and I couldn't find those variables.

ebebbington:
> By "GITHUB OWNER" and "GITHUB_BRANCH", are you talking about environment variables? I checked Github's [environment variables](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables), and I couldn't find those variables.

Yeeah it was more pseudo code, i'm not sure how we'd do that bit tho, i need to think it through

ebebbington:
@MVEMCJSUNPE could you also add `console.log(Deno.env.get("GITHUB_OWNER")` too if that's ok? Sorry if i'm bugging you

I'm thinking we could do:
```
const url = `${Deno.env.get("GITHUB_OWNER")}/deno-drash/${Deno.env.get("GITHUB_REPOSITORY")}`
```

MVEMCJSUNPE:
I think BASE_REF looks pretty promising. Maybe I'll just use that for latestBranch.

ebebbington:
GITHUB_BASE_REF would always be master i believe, see [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#:~:text=Environment%20variables%20are%20case%2Dsensitive,entire%20workflow%20using%20the%20jobs.)

MVEMCJSUNPE:
I think we should just accept that there's no way to get the repository ("MVEMCJSUNPE/deno-drash") of whoever forked from the main repository (drashland/deno-drash). Since this is the case, the environment variable "GITHUB_HEAD_REF" won't ever work for forked repos.

ebebbington:
So i guess, all the getting of the repo logic could just be:
```
// supports forks
let drashUrl = "https://raw.githubusercontent.com/" + Deno.env.get("GITHUB_OWNER") +
  `/deno-drash/${latestBranch}`; // https://raw.githubusercontent.com/<NAME>/deno-drash/<branch>

// if fork doesnt exist, use drashland repo, eg name might be ebebbington, but i dont have a fork
try {
  const res = await fetch(drashurl)
  await res.json()
  if (res.status !== 200) {
    drashUrl = "https://raw.githubusercontent.com/drashland" +
  `/deno-drash/${latestBranch}
} catch (err) {
  // do nothing
}
```

crookse:
i'm going to investigate this as well

ebebbington:
@MVEMCJSUNPE using your help, i managed to get it working :D #410

MVEMCJSUNPE:
How'd I help?

MVEMCJSUNPE:
I got it to work again locally, but the CI tests just keep saying no.

PR #412, opened 2020-10-29
feat: Throw an error if response is not correctly returned from resource

### LINKED ISSUE #402: Throw 418 HTTP error if resources don't return responses
## Summary

What: `Server.handleHttpRequest()` should throw 418 errors when a resource doesn't return a response object.

Why: If a resource doesn't return a response object, the request-resource-response lifecycle will continue to process and will try to get `response.status_code` even if there is no `response` object returned. This will throw the following error:

```
Cannot read property 'status_code' of undefined
```

This error message isn't helpful when it comes to debugging. This error should be handled better.

## Acceptance Criteria

- [ ] Throw the following error if a resource doesn't return a response:
    ```typescript
    throw new Drash.Exceptions.HttpResponseException(418, `Response was not returned from ${resourceClass.name}.${request.method}().`);
    ```
- [ ] Tests

ebebbington:
Fixes #402 

**Description**

* If the data returned from a method is not of type Response or ResponseOutput, throw an error

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -24,11 +24,12 @@ interface IRequestOptions {
  * Used to check if a response object is of type Drash.Interfaces.ResponseOutput
  */
 // @ts-ignore Only exception because response cannot be properly typed and we're checking if it is of type interface anyway
-function responseIsOfTypeResponseOutput (response: any): boolean {
+function responseIsOfTypeResponseOutput(response: any): boolean {
   if (typeof response === "object" && Array.isArray(response) === false) {
-    return "status" in response && "headers" in response && "body" in response && "send" in response && "status_code" in response
+    return "status" in response && "headers" in response &&
+      "body" in response && "send" in response && "status_code" in response;
   }
-  return false
+  return false;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


ebebbington:
ok ready for review

src/http/server.ts
```diff
@@ -20,6 +20,18 @@ interface IRequestOptions {
   };
 }
 
+/**
+ * Used to check if a response object is of type Drash.Interfaces.ResponseOutput
+ */
+// @ts-ignore Only exception because response cannot be properly typed and we're checking if it is of type interface anyway
+function responseIsOfTypeResponseOutput(response: any): boolean {
+  if (typeof response === "object" && Array.isArray(response) === false) {
```
crookse, line 28: would it every be an array? i ask because our docs explicitly state to `return this.response`. although this does cover if someone wants to try returning an array.

src/http/server.ts
```diff
@@ -257,7 +269,19 @@ export class Server {
       if (typeof resource[request.method.toUpperCase()] !== "function") {
         throw new Drash.Exceptions.HttpException(405);
       }
-      response = await resource[request.method.toUpperCase()]();
+      response = await resource[request.method.toUpperCase()](); // response can  be literally anything, it's down to the user what they return from the method
```
crookse, line 272: can you put the comment above this line? (nitpicky i know... sorry!)

src/http/server.ts
```diff
@@ -257,7 +269,19 @@ export class Server {
...
+
+      // Check the response was returned as the Drash.Http.Response type, or as ResponseOutput
+      const isValidResponse =
```
crookse, line 275: we should turn this into a function and make it testable

src/http/server.ts
ebebbington: This is more a specific edge case, because `typeof [] === "object"` is `true`,  we're making sure that it is a proper object so there are no uncaught errors

src/http/server.ts
ebebbington: good shout

src/http/server.ts
ebebbington: for example, `response` could be `"hello"`, and i thought it was best to add the check inside the function so we can do `responseIsOfTypeObject(...)` wherever we want without adding code each time to check if its a proper object

src/http/server.ts
ebebbington: https://github.com/drashland/deno-drash/pull/412/commits/18a790908aae005f3c483cecf3bdb04349a42958

src/http/server.ts
ebebbington: https://github.com/drashland/deno-drash/pull/412/commits/18a790908aae005f3c483cecf3bdb04349a42958

src/http/server.ts
crookse: cool!

crookse (approved):

Issue #413, opened 2020-10-30
Compile time level middleware


crookse:
## Dependents

https://github.com/drashland/deno-drash-middleware/issues/59

The above issue cannot be released until this issue issue is closed and it's code is released.

## Summary

What: Add compile-time middleware to be process data at compile time to be used at runtime.

Why: Drash has server-level middleware and resource-level middleware. Server-level and resource-level middleware are executed during Drash's runtime. It would be nice to be able to process data during compile time and use that data during runtime. For example, if I want to compile some data to be used at runtime, but I don't want that data to be compiled during runtime (for performance reasons), it would be nice to compile the data when the server is being instantiated. That way the data is ready to be used during runtime -- keeping the performance up. A use case for this is compiling caches during compile time and using whatever is in the cache during runtime.

## Acceptance Criteria

- [x] Add logic that processes compiled data from compile time level middleware during runtime
- [x] Documentation (https://github.com/drashland/website/issues/210)
- [x] Update Drash Lifecycle Diagram (drashland/website#229)
- [x] Tests

PR #414, opened 2020-10-30
feat: Issue #413 compile time middleware

crookse:
Addresses #413 

## Description

* Adds logic to execute compile time middleware. When compile time middleware is executed, runtime middleware gets set to run at a later time (only after requests -- i don't see a reason from runtime data to be executed before a request).
* https://github.com/drashland/deno-drash-middleware/issues/59 is dependent on this feature.


crookse:
closing.. changing branch name to fix tests

PR #419, opened 2020-10-31
fix: optional args regex not matching non-alphanumeric characters

### LINKED ISSUE #370: Optional args in variable paths are not parsing correctly with non-alphanumeric values
**Describe the bug**
When a resource path is defined with an optional arg it will not match if the trailing arg value includes a non-alphanumeric character:

**To Reproduce**

Given the resource:
```
export class SomeResource extends Drash.Http.Resource {
  static paths = ["/:username/:slug?"];
  public GET() {
    let username = this.request.getPathParam("username");
    let slug = this.request.getPathParam("slug");
    this.response.body = `${username}'s stuff`;
    if(slug) this.response.body += ` - ${slug}`;
    return this.response;
  }
}
```

Go to the URL path: /userx/test123 => Works
Go to the URL path: /userx/test-123 => Not found

Change path to:

```
static paths = ["/:username/:slug?/:empy?"];
```

Go to the URL path: /userx/test123 => Works
Go to the URL path: /userx/test-123 => Found but: slug = "test"

**Expected behavior**
It should match and return the full value

**Suggested Solution(s)**
Fix parsing

crookse:
Fixes #370 

**Description**

* Removes hardcoded alphanumeric matching on first param which caused the regex to stop matching early on

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -700,7 +700,7 @@ export class Server {
             // global, and accounts for there being a required parameter before
             tmpPath = tmpPath.replace(
               /\/(:[^(/]+|{[^0-9][^}]*}\?)\/?/,
-              "/?([a-zA-Z0-9]+)?/?", // A `/` being optional, as well as the param being optional, and a ending `/` being optional
+              "/?([^/]+)?/?", // A `/` being optional, as well as the param being optional, and a ending `/` being optional
             );
           } else { // We can now create the replace regex for the rest taking into consideration the above replace regex
             tmpPath = tmpPath.replace(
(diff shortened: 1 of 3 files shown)


ebebbington:
merge master in to fix the http tests

ebebbington:
oh lmao i know why, rawgithub doesn't like hashes in the branch name (url)

crookse:
> oh lmao i know why, rawgithub doesn't like hashes in the branch name (url)

HAHA wow.

ebebbington:
> > oh lmao i know why, rawgithub doesn't like hashes in the branch name (url)
> 
> HAHA wow.

Lol, yeah... usually i just close the pr, and open a new one without using the # in the branch name, im sure there's a better way tho

crookse:
closing because of failing tests.. need to take the # out of the branch name

PR #420, opened 2020-10-31
fix: optional args regex not matching non-alphanumeric characters

### LINKED ISSUE #370: Optional args in variable paths are not parsing correctly with non-alphanumeric values

crookse:
Fixes #370 

**Description**

* Removes hardcoded alphanumeric matching on first param which caused the regex to stop matching early on

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -700,7 +700,7 @@ export class Server {
             // global, and accounts for there being a required parameter before
             tmpPath = tmpPath.replace(
               /\/(:[^(/]+|{[^0-9][^}]*}\?)\/?/,
-              "/?([a-zA-Z0-9]+)?/?", // A `/` being optional, as well as the param being optional, and a ending `/` being optional
+              "/?([^/]+)?/?", // A `/` being optional, as well as the param being optional, and a ending `/` being optional
             );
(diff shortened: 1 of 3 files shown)


ebebbington (approved):
LGTM, i see you added the edge case into the test 👍

PR #421, opened 2020-10-31
feat: compile time middleware

crookse:
Addresses #413 

## Description

* Adds logic to execute compile time middleware. When compile time middleware is executed, runtime middleware gets set to run at a later time (only after requests -- i don't see a reason from runtime data to be executed before a request).
* https://github.com/drashland/deno-drash-middleware/issues/59 is dependent on this feature.


src/http/server.ts
```diff
@@ -471,9 +482,26 @@ export class Server {
       // requested asset. Since this occurs after the MIME code above, the
       // client should receive a proper response in the proper format.
       if (!this.configs.pretty_links || request.url.split(".")[1]) {
-        response.body = Deno.readFileSync(`${this.directory}/${request.url}`);
-        await this.executeMiddlewareServerLevelAfterRequest(request, response);
-        return response.sendStatic();
+        try {
+          // Try to reaed the file if it exists
```
ebebbington, line 486: *reaed*

src/http/server.ts
```diff
@@ -851,15 +892,56 @@ export class Server {
...
+   * @param response - The response object.
+   *
+   * @returns A Drash response or a boolean. If it returns a boolean, it means
+   * it did or did not process properly (true being it processed; false being it
+   * did not process).
+   */
+  protected async executeMiddlewareServerLevelRuntimeAfterRequest(
+    request: Drash.Http.Request,
+    response: Drash.Http.Response,
+  ): Promise<void> {
```
ebebbington, line 927: Doc block says it returns a drash response or boolean, but method return type is void

src/http/server.ts
```diff
@@ -303,6 +313,7 @@ export class Server {
         throw new Drash.Exceptions.HttpResponseException(
           418,
           "The response must be returned inside the " +
+            resource.constructor.name + "." +
             request.method.toUpperCase() + " method of the resource",
```
ebebbington, line 317: This results in `The response must be returned inside the HomeResource.GET method of the resource`, wonder if it sounds better to you for it to be `The response must be returned inside the GET method of the "HomeResource"`?

ebebbington (changes requested):
Code looks good to me, a few comments, but i'm still unsure how a user would go about using ts on the client without transpiling and using compile time middleware?

src/http/server.ts
crookse: that makes more sense

crookse: 0aa19c1

src/http/server.ts
crookse: 93082a1 (it supposed to return void -- i removed the @return part)

src/http/server.ts
crookse: 7c1d0c8 (ok it READs better now)

ebebbington (approved):
image

Issue #399, opened 2020-10-23
Deprecate template engine logic


crookse:
## Summary

What: Deprecate template engine logic.

Why: Template engine logic is not part of the core processes of Drash. Template engine logic will be written in the deno-drash-middleware repo.

**REMOVAL DATE: January 1, 2021**

## Acceptance Criteria

- [x] Deprecate any code using the template engine
- [x] Notify users of deprecation date. We can use the `LoggerService` to log the following in the methods that use deprecated code:
    ```
    WARN You are using deprecated code. See migration solutions here: https://github.com/drashland/deno-drash/issues/xxx
    ```
- [x] ~Keep this issue open until code has been removed in a future release~ Making a new issue and pinning it.
- [x] ~Delete code on deprecation date (a PR should be made that we just merge in on the deprecation date)~ Making a new issue and pinning it.
- [x] Write migration guide (https://github.com/drashland/website/issues/221)
- [x] Implement template engine middleware (https://github.com/drashland/deno-drash-middleware/issues/53)

MVEMCJSUNPE:
Does this include integration tests?

crookse:
@MVEMCJSUNPE, yup!

MVEMCJSUNPE:
@crookse Can I work on this issue?

crookse:
@MVEMCJSUNPE be our guest :)

crookse:
i assigned it to you, @MVEMCJSUNPE

MVEMCJSUNPE:
Thanks.

PR #424, opened 2020-10-31
deprecate: Drash.Compilers.TemplateEngine

MVEMCJSUNPE:
Attempts to fix [ #399 ]
**Description**
I used the method addLogger to add a ConsoleLogger 'DeprecationLogger'
to the Loggers namespace in the constructor of TemplateEngine. To me,
this was the most logical place to call addLogger. Then, log messages were
added at any instance TemplateEngine was used. I don't see
why we couldn't just add a deprecation log message at the beginning of
the constructor for TemplateEngine.

**Notes**
* Since the deprecation date is TBD, I don't have that in any of the log messages YET. Maybe in another commit, it will be added


MVEMCJSUNPE:
Whoops. Maybe addLogger is only supposed to be used in certain situations?

crookse:
add logger was intended to be used by external parties. we have an internal logging service called the LoggerService located here: https://github.com/drashland/services/tree/master/logger

i should've been more explicit in the issue (my apologies). this issue should be using the LoggerService though (exporting it from the deps file and importing it where needed -- in this case the template engine file). and then you would do a LoggerService.warn() call with a message.

the removal date is January 1, 2021. i just realized we can't merge this code until we have a migration guide written. because if we state there's a migration guide in the warning message and we don't have one yet, then yea... that'd be bad lol. i'll work on a migration guide tomorrow though so we can push this item out. i really want this code stripped out and placed into middleware. it only convolutes the codebase.

anyways, we REALLY appreciate this work :) i'll let you know when i've done the migration guide

MVEMCJSUNPE:
Oh, sorry, my bad. I looked through the code base for an example usage of LoggerService and couldn't find one. So I did some Googling, and I saw that [LoggerService](https://docs.nestjs.com/techniques/logger) for nest js seemed to implement a lot of the same methods; from there, I just assumed that Logger from logger.ts was the LoggerService that was being referred to, though I kind of had a feeling that something was amiss.

crookse:
@MVEMCJSUNPE, i just finished developing the middleware features to help move this issue along. they'll need to be reviewed and closed out before we can move this PR along. i'm thinking we'll have this PR merged in by the end of the week though. also, your changes look REALLY good. thanks! there will be one last change to the warning message. i'm going to be finishing the migration guide maybe tonight or tomorrow. then you can update the warning message to point users to the new migration guide link.

MVEMCJSUNPE:
Alright, I'll keep an eye out.

crookse:
Alright I wrote documentation on the new template engine middleware. The warning message in this PR should read as follows:

```
You are using deprecated code, which will be removed on January 1, 2021. Please update your application to use Tengine -- a template engine middleware. View it at https://github.com/drashland/deno-drash-middleware/tengine.
```

@ebebbington @Guergeiro @saragee3, we'll need to do all of the following on the same day:

* Merge #428 (after_resource hook)
* Merge this PR
* Release deno-drash v1.3.0
* Merge https://github.com/drashland/deno-drash-middleware/pull/62 (Tengine - A template engine middleware)
* Release deno-drash-middleware v0.6.1

I'm pinning the issue associated with this PR so that we can be reminded it needs to be addressed on January 1, 2021.

MVEMCJSUNPE:
So just to be clear, I only have to log deprecation warnings in every method in the class TemplateEngine in src/compilers/template_engine.ts, right? Are there any other places I should log a deprecation warning? In the original issue, it says "Deprecate any code using the template engine", which one could interpret as "For any code that declares a new instance of TemplateEngine from src/compilers/template_engine.ts, add a @deprecated JS comment, and a deprecation log message before it". So this would include all the files listed in the picture: 
template_engine_instances

crookse:
@MVEMCJSUNPE, sorry for the confusion and you are correct. the acceptance criteria says all methods, but it should only be in the template engine's constructor. thanks for bringing this up.

crookse (approved):

PR #428, opened 2020-11-02
feat: after_resource middleware hook

### LINKED ISSUE #427: after_resource middleware hook
## Summary

What:

In the request-resource-response lifecycle, we need to execute a hook that occurs right before the resource executes it's HTTP method. This hook will execute AFTER the resource has been found. Therefore, the name will be `after_resource`. Although this name might change in v2 (see #426 for that discussion).

Why:

We're in the process of ripping out the template engine logic. It's going to be deprecated soon and removed on January 1, 2020. In order to use a template engine as middleware, the middleware would need to override the `response.render()`, method; and as of right now there is no way to do that. We need to add an `after_resource` middleware hook to the request-resource-response lifecycle to accomplish this.

## Acceptance Criteria

- [x] Tests
- [x] Add middleware hook after a resource is found, but before the resource's HTTP method is executed
- [x] Update Drash Lifecycle Diagram (https://github.com/drashland/website/issues/229)
- [x] Documentation on using after_resource middleware (https://github.com/drashland/website/issues/230)

crookse:
Closes #427 

**Description**

* Add middleware hook after a resource is found, but before the resource's HTTP method is executed
* Makes `response.render()` return `Promise<boolean|string>` and `boolean | string`.
* This is a prerequisite for template engine middleware.
* https://github.com/drashland/deno-drash-middleware/issues/53 is dependent on this code being released.
* https://github.com/drashland/deno-drash/issues/399 is dependent on this code being released.

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -124,9 +124,9 @@ export class Response {
    * @returns The html content of the view, or false if the `views_path` is not
    * set.
    */
-  public render(
+  public async render(
     ...args: unknown[]
-  ): string | boolean {
+  ): Promise<boolean|string> {
     if (!this.views_path) {
       return false;
     }
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -126,7 +126,7 @@ export class Response {
    */
   public render(
     ...args: unknown[]
-  ): boolean|string {
+  ): Promise<boolean|string> | boolean | string {
     if (!this.views_path) {
       return false;
     }
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -126,7 +126,7 @@ export class Response {
    */
   public render(
     ...args: unknown[]
-  ): Promise<boolean|string> | boolean | string {
+  ): Promise<boolean | string> | boolean | string {
     if (!this.views_path) {
       return false;
     }
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -124,9 +124,9 @@ export class Response {
    * @returns The html content of the view, or false if the `views_path` is not
    * set.
    */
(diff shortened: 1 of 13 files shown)


src/http/response.ts
```diff
@@ -126,7 +126,7 @@ export class Response {
    */
   public render(
     ...args: unknown[]
-  ): string | boolean {
+  ): Promise<boolean | string> | boolean | string {
```
ebebbington, line 129: Breaking change i guess?

Anyone using render now will fall into a race condition where the response is returned in the resource method before the body is set?

ebebbington (approved):
image

Just one comment

src/http/response.ts
crookse: the tests right now show that it works as it did before. i can add async tests to make sure though

ebebbington: i cant see any awaits inside the `render` function?

crookse: yea there aren't any. it's to handle template engines that use `await`. for example, this has `Promise<boolean|string>` now because of Eta. Eta requires `await renderFile()`. when people use Eta, they're going to define the render method in Tengine and call `await renderFile()`, that's where Drash's `render()` method needs to return a `Promise`.

crookse: what `Tengine` will do is replace Drash's `render()` method with Eta's `await renderFile()` method. this is the same for any other template engine if they use `await`.

ebebbington: Ah yeah gotcha

ebebbington:
fyi im happy with this pr

PR #445, opened 2020-11-07
Refactor create app

### LINKED ISSUE #437: Clean up create_app.ts script tests
## Summary

What:

Most of the assertions occur in one test case. They should be broken apart into their own test cases.

Why:

We should break out these assertions into their own test cases so that we can debug more easily and faster.

## Acceptance Criteria

- [ ] Refactor create_app.ts script unit tests
- [ ] Tests pass

## Taking on this issue? Read this before you get started.

If you take on this issue and the acceptance criteria does not make sense, post comments on this issue with your questions. Also, you can always hop in the Discord channel at https://discord.gg/c5R3sYuR to get help quickly.

Please read the [Contributing Guidelines](https://github.com/drashland/.github/blob/master/CONTRIBUTING.md) before starting this issue.

MVEMCJSUNPE:
Attempts to fix #437 

## Summary

* Tried automating test cases through for loop and generator

## Other Notes
There's only one failure in create_app_local, and about a dozen or so in create_app_http. I'm not sure what's going on in create_app_http at the moment.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,0 @@
-Fixes [#<issue number>](url of issue)
-
-**Description**
-
-* Bullet points of developments made
-
-**Other Notes**
-
-* Any other useful information goes here
--- a/.github/workflows/bumper.yml
+++ b/.github/workflows/bumper.yml
@@ -7,7 +7,7 @@ jobs:
   update-dep:
     strategy:
       matrix:
-        deno: ["1.5.1"]
+        deno: ["1.5.2"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -39,11 +39,11 @@ jobs:
       - name: Create App (local)
         run: |
           deno cache create_app.ts
-          deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env
+          deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env --unstable
 
       - name: Create App (http)
         run: |
-          deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
+          deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net --unstable
 
   benchmarks:
     strategy:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        deno: ["1.5.1"]
+        deno: ["1.5.2"]
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -49,7 +49,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        deno: ["1.5.1"]
+        deno: ["1.5.2"]
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -39,11 +39,11 @@ jobs:
       - name: Create App (local)
         run: |
           deno cache create_app.ts
-          deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env --unstable
+          deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env
 
       - name: Create App (http)
         run: |
-          deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net --unstable
+          deno test tests/cli/create_app_test_http.ts --allow-read --allow-write --allow-run --allow-env --allow-net
 
   benchmarks:
     strategy:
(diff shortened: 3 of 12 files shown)


MVEMCJSUNPE:
This is a WIP btw. I'm not sure how to add that label.

ebebbington:
Did it for you :)

ebebbington:
The tests are failing because the other test commands will need the `--unstable` flag too, but I don't think we'd want to go with any unstable modules - i know a way you can check if a file exists without the unstable fs module: https://stackoverflow.com/a/61868755/11594164
```

**Comment by MVEMCJSUNPE** [comment]
My bad. I thought that it would be better to reuse code rather than writing our own implementation (fileExists). I personally don't think existsSync should be unstable. I looked at the source code for existsSync, and it uses Deno.lstatSync, and according to the [manual](https://deno.land/manual@v1.0.0/runtime/stability), Deno.lstat should be safe to use. I'll re-add fileExists in another commit.

**Comment by ebebbington** [comment]
Oh that's interesting... as a side note, might be interesting to file and issue with deno to remove the unstable flag for it, and see what they say?

**Comment by MVEMCJSUNPE** [comment]
I readded fileExists and fixed the errors in create_app_test_http.ts. I've got duplicate copies of add_suffix and FilePairGenerator in create_app_test_local.ts and create_app_test_http.ts. Should I place those functions in the deps.ts that is in the same directory  as create_app_test_local.ts and create_app_test_http.ts or the deps.ts located one directory above? Or just leave those functions be?

**Comment by MVEMCJSUNPE** [comment]
Looking at this [reddit post](https://www.reddit.com/r/Deno/comments/ib2ho3/deno_std_library_stable_release_date/), I think the Deno std library won't become stable anytime soon. They probably always consider the question of when to make it stable, so I'm not going to bother.

**Comment by ebebbington** [comment]
Ah yes ignore me about the stable thing, I’m an idiot xD I forgot it was part of std for some reason...

**Comment by MVEMCJSUNPE** [comment]
The tests pass by the way. @crookse or @ebebbington , feel free to take a look.

**Comment by crookse** [comment]
hey @MVEMCJSUNPE, i hate to do this to you, but there's a lot of code bloat here imo. i like automation, but not in this case. what i was expecting was that each assertion would be wrapped in a Rhum.testCase() block -- essentially allowing us to find out exactly where an assertion failed. the for loops you added kind of accomplish this (and correct me if i'm wrong), but i don't think, if an assertion were to fail, that the assertion error would tell us exactly what assertion failed. it would tell us that an assertion in the loop failed, but it wouldn't tell us anything further and we'd have to sift through code to figure out what exactly triggered the failure.

**Comment by MVEMCJSUNPE** [comment]
I'm not sure about code bloat; if you're talking about code duplication and the violation of the DRY ([Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle), then I completely agree, although the original create_app tests also were very 'repeat-y'. However, I'm not thoroughly convinced that my edits result necessarily result in a loss of information that would be useful for debugging. So if I make a small change to the create_app.ts file, changing 
https://github.com/drashland/deno-drash/blob/98f81768cef743ed98cc64c912325d48871808cd/create_app.ts#L210-L213

to:

```ts
    await copyFile(
      "/tests/resources/home_resource_test_web_app.ts",
      "/tests/resources/home_resource_test_web_app.ts",
    );
```

, then obviously, one or more of the tests is going to fail. If we look at the error output from running `deno test tests/cli/create_app_test_local.ts --allow-read --allow-write --allow-run --allow-env`, we get the following output: 
create_app_test_screen1
create_app_test_screen2
create_app_test_screen3

So, if we pretend that we had no idea why the test was failing and got out our rubber ducky for debugging, the first thing we would notice is that the error is a result of tests/resources/home_resource_test.ts not being created correctly. And if we take a look at the line where it failed, which deno generously gives in the error output, we see: 
create_app_code
so the error is a result of tests/resources/home_resource_test.ts not existing. And I would argue that this is the exact amount of information the the original create_app tests gave. If we checkout the last commit that was made before I made any changes (d9f15b42843fd3961f5b998a0eba6fe2462b1f70), and we rerun create_app_test_local.ts, Deno gives us the following output: 
old_create_app_test_screen1
old_create_app_test_screen2
So we get essentially the same output from Deno. If we look at the lines of code around where Deno hinted there was an error in the stack trace, we see: 
create_app_old_code
So, yes, one might argue that the file that is being copied from is given in the old code, and that would be a point in your favor. However, that could find that information in the new code by scrolling to the top of the test suite:
create_app_suffixdict

So in conclusion, since programmers receive the same amount of information from the new version as the old version of the create_app tests, the old version is just as difficult to debug as the new version should any errors occur.

MVEMCJSUNPE:
@crookse but if you still think I should change the create_app tests code after I've made my case, then I'll still do it.

crookse:
i get where you're coming from, but why would i want to look around the code (or scroll up) if i don't have to? i can't have this code merged knowing i'm going to have to figure out how it works again later in the future. i dont think this PR solves the issue (and it may be that i should've been more explicit and provided psuedocode). i'm picky about the code that goes in drash and i'm going to have to punt on this solution.

MVEMCJSUNPE:
I got rid of the for loops and did the test cases the way people normally do.

ebebbington (commented):
LGTM, thanks again @MVEMCJSUNPE :) Also just wanted to say, we appreciate all the input and issues you have tackled :)

Commenting because will need another approval

MVEMCJSUNPE:
Thanks @ebebbington .

crookse (approved):

PR #446, opened 2020-11-08
chore: Improve readability of code

ebebbington:
## Summary

* Improve readability of `server.handleHttpRequest()`
* Improve readability of `server.constructor()`
* Improve semantics of method names
* Breaks apart `server.handleHttpRequest()` into smaller functions (these functions are protected)
* Breaks apart `server.constructor()` into smaller functions (these functions are protected)
* Alphabetizes methods

## Examples

`server.constructor()` now looks like:

```typescript
  constructor(configs: Drash.Interfaces.ServerConfigs) {
    this.configs = this.buildConfigs(configs);

    this.addMiddleware();

    this.addResources();

    this.addStaticPaths();

    this.addTemplateEngine();
  }
```

`server.handleHttpRequest()` method now looks like:

```typescript
  public async handleHttpRequest(
    serverRequest: ServerRequest,
  ): Promise<Drash.Interfaces.ResponseOutput> {
    const request = await this.buildRequest(serverRequest);
    let response = this.buildResponse(request);

    try {
      await this.executeMiddlewareBeforeRequest(request);

      // Is this request for the favicon?
      if (request.url == "/favicon.ico") {
        return await this.handleHttpRequestForFavicon(request);
      }

      // Is this request for a static asset?
      if (this.requestTargetsStaticPath(request)) {
        return await this.handleHttpRequestForStaticPathAsset(request);
      }

      // Is this request for a virtual path asset?
      if (this.requestTargetsVirtualPath(request)) {
        return await this.handleHttpRequestForVirtualPathAsset(request);
      }

      // If all conditions above fail, then a resource is likely being requested
      return await this.handleHttpRequestForResource(request, response);
    } catch (error) {
      return await this.handleHttpRequestError(
        request,
        new Drash.Exceptions.HttpException(error.code ?? 400, error.message)
      );
    }
  }

```


ebebbington:
might be easier to take a look at [server.ts](./src/http/server.ts)

crookse:
@Guergeiro / @saragee3,

i suggest you pull down the branch and look at how server.ts is now written. the diff makes it difficult to read. the important parts are that the `constructor` is now cleaned up and `handleHttpRequest()` is now cleaned up

src/http/server.ts
```diff
@@ -437,44 +250,63 @@ View more information regarding this deprecation/removal at:
...
-      response.body = "";
     }
 
-    let output: Drash.Interfaces.ResponseOutput = {
-      body: response.body as string,
-      headers: headers,
-      status: response.status_code,
-    };
+    return response.send();
+  }
```
Guergeiro, line 271: There is no asynchronous code in this function... Now, instead of making it a normal function, I would keep the `async` signature and add asynchronous code (it's always better). How? Make the file access stuff using their asynchronous methods.
```ts
const path = `${await Deno.realPath(".")}/favicon.ico`;
response.body = await Deno.readFile(path)
```

Guergeiro (changes requested):
This is something we need to avoid doing in the future. Sync stuff blocks I/O, so let's avoid using.

src/http/server.ts
Guergeiro: Actually, this happens in a bunch of places. Here's a list:

- handleHttpRequestForStaticPathAsset
- handleHttpRequestForVirtualPathAsset

crookse: 6a35af7

Guergeiro (dismissed):
Seems good to me 👍🏿

.github/workflows/release.yml
```diff
@@ -6,10 +6,10 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: denolib/setup-deno@master
-      - run: deno install -A -f --unstable -n eggs https://x.nest.land/eggs@0.3.2/eggs.ts
-      - run: |
+     - uses: actions/checkout@v2
+     - uses: denolib/setup-deno@master
+     - run: deno install -A -f --unstable -n eggs https://x.nest.land/eggs@0.3.2/eggs.ts
+     - run: |
```
ebebbington, line 12: i need to fix this

crookse: good catch

ebebbington: https://github.com/drashland/deno-drash/pull/446/commits/601b33a9b91ae7085bce83a40b310d8224a53187

ebebbington:
What are we doing with this pr?

crookse:
waiting for @saragee3 to review it

saragee3 (dismissed):
yowza!

PR #455, opened 2020-11-22
chore: Remove bloat

ebebbington:
## Summary

* Uses latest deno version in workflows
* Cleans up bumper file following the above
* Remove duplication in README


crookse (approved):

ebebbington:
LGTM, you cool with this eric?

crookse:
yup. i removed more stuff from the readme -- essentially trying to direct the user to our docs. good call on removing a lot of this duplication btw. it made me realize we're creating two "sources of truth" when we should only have one.

PR #462, opened 2020-12-13
chore: fix failing tests

crookse:
Not sure why the tests were passing for anyone else, but I've sifted through the code and found some areas where they should've failed.

This is especially true for the `file_logger_test.ts` file. It was recreating the `tmp` folder. I already have a `tmp` folder where I do manual testing of the framework. The test would try to make the `tmp` folder and delete the `tmp` folder. It was failing to do so and causing errors.


ebebbington (approved):
LGTM

PR #463, opened 2020-12-19
fix: create app http tests

crookse:
## Summary

* makes sure `master` is set if `Deno.env.get("GITHUB_HEAD_REF")` returns a falsy value or `""`.


ebebbington (approved):
Yaaaaaay nice one

Guergeiro (approved):
Oh boy, JavaScript being JavaScript 👍🏿

PR #482, opened 2021-03-15
chore: add script to generate many resources to aid in benchmarking

ebebbington:
Essentially just saves this script so we don't have to manually write it again


console/make_resources.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Purpose of this script:
+ *
+ *   Create a set number of resources for a Drash server to use,
+ *   to benchmark how Drash will perform when hundreds or thousands of
```
saragee3, line 5: i'm curious why this is needed? it seems like an anti-pattern and whoever's app should probably be rearchitected if it has thousands of resources. will this test case ever be in the real world?

ebebbington: Well maybe not, but i guess the purpose of it is just for us to quickly benchmark a drash app against more than once resource (ie a more realworld demonstration than the basic benchmark file we already have that only has one **very basic resource**)

If you so feel it really isn't needed then that's fine, we can close this PR, i guess this was merely to "reference" the code to do this, to save me or eric rewriting it again - but there's nothing stopping me (or him) just saving it as a Gist or into a random repo for our **personal** GH accounts

saragee3: yeah, it feels like it's in the wrong repo ... maybe services or castle? BUT ultimately you guys decide :) @crookse

PR #485, opened 2021-03-29
chore: Notify middleware repo on new drash releases

ebebbington:
## Summary

Part 1 of 2

* Sends an event to the middleware repo on a release. This will be used by the middleware repo to bump it's dependencies, and hopefully pre-release (part 2)
* Why? We found that to make middleware usable, the drash version it uses has to be the same as the one a user uses, so say a user is using latest of drash (1.4.2), and uses middleare too, but the middleware uses drash v1.4.1, this is incompatible and will throw an error. So with this PR, we're making sure that when drash releases a new version, we release a new middleware version right away that uses the latest of drash

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,3 +14,15 @@ jobs:
             -H "Content-Type: application/json" \
             --data '{"event_type": "release", "client_payload": { "repo": "deno-drash", "module": "drash", "version": "${{ github.ref }}" }}' \
             https://api.github.com/repos/drashland/castle/dispatches
+
+  update-middleware-repo:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Middleware about this release
+        run: |
+          curl -X POST \
+            -u "${{ secrets.CI_USER_NAME }}:${{ secrets.CI_USER_PAT }}" \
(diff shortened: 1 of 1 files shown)


saragee3 (approved):

Guergeiro:
The bumper scripts already run on the middleware repo (where Drash is a dependency), so why do we need yet another bumper?

.github/workflows/release.yml
```diff
@@ -14,3 +14,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Middleware about this release
+        run: |
+          curl -X POST \
+            -u "${{ secrets.CI_USER_NAME }}:${{ secrets.CI_USER_PAT }}" \
+            -H "Accept: application/vnd.github.everest-preview+json" \
+            -H "Content-Type: application/json" \
+            --data '{"event_type": "release", "client_payload": { "repo": "deno-drash", "module": "drash" }}' \
+            https://api.github.com/repos/drashland/deno-drash-middleware/dispatches
```
crookse, line 28: we need to catch this event in middleware right?

ebebbington: Yeah that’s gonna be done as a separate PR

PR #492, opened 2021-04-19
feat: support send binary body

### LINKED ISSUE #491: How to send binary body?
## Summary
Such as responding to a document?
```
this.response.body = (await Deno.readFile("test.png"));
```
This will get a null.

juzi5201314:
Fixes #491

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -119,7 +119,7 @@ export class Response {
    *
    * @returns The response in string form.
    */
-  public generateResponse(): string {
+  public generateResponse(): string | Uint8Array {
     let contentType = this.headers.get("Content-Type");
 
     switch (contentType) {
@@ -138,6 +138,9 @@ export class Response {
         }
         if (typeof this.body === "boolean") {
           return this.body.toString();
+        }
+        if (this.body instanceof Uint8Array) {
+          return this.body;
         }
         if (typeof this.body !== "string") {
           // final catch all, respond with a generic value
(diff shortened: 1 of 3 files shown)


ebebbington:
Hi @juzi5201314, ideally a unit and integration test would be written to ensure that, if a resource does return a typed array, that the response given is correct (the data is intact)

And I think a better approach would be to add onto the `generateResponse()` method in the Response class, such as adding another conditional for if the body is an instance of a Uint8Array, then on top of this, inside the `send()` method, only encode if the body isn’t already encoded (eg a typed array)

juzi5201314:
> Hi @juzi5201314, ideally a unit and integration test would be written to ensure that, if a resource does return a typed array, that the response given is correct (the data is intact)
> 
> And I think a better approach would be to add onto the `generateResponse()` method in the Response class, such as adding another conditional for if the body is an instance of a Uint8Array, then on top of this, inside the `send()` method, only encode if the body isn’t already encoded (eg a typed array)

good idea

crookse:
hey @juzi5201314, sorry i haven't been around to investigate, but i did investigate this a bit like 10 minutes ago and this PR is on the right track. could you look into why the tests are failing please? if you need help, feel free to reach out here on github or in our discord. we'll be happy to help :)

juzi5201314:
> hey @juzi5201314, sorry i haven't been around to investigate, but i did investigate this a bit like 10 minutes ago and this PR is on the right track. could you look into why the tests are failing please? if you need help, feel free to reach out here on github or in our discord. we'll be happy to help :)

I think I already know why an error occurred in ci and I didn't have it when I tested it locally, because my local deno is not the latest main branch, and I cloned the deno of the main branch and tested it, the same error occurred!
But I am still looking for the specific cause of the error in deno.

juzi5201314:
I found the change that caused the drash test error ([fe59e7ae6034249c47eaf94f4258bb4467153f0d](https://github.com/denoland/deno/commit/fe59e7ae6034249c47eaf94f4258bb4467153f0d)) through the elimination method.
I will try to fix it next

crookse:
this is good by me. @ebebbington ?

crookse (dismissed):
tests are failing because we need to fix our integration tests, not because of this PR. approved

ebebbington (approved):
Looks good to me :)

PR #503, opened 2021-06-13
chore: Support before request middleware wanting to end req lifecycle

ebebbington:
Address the issue found here: drashland/deno-drash-middleware#97 (comment)

I tested a local version of RateLimit and it worked fine

I also believe in the middleware functions, they could just call "response.send()" if they want to end the request lifecycle, as opposed to drash adding logic to control this

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -140,7 +140,7 @@ export class Server {
     let response = this.buildResponse(request);
 
     try {
-      await this.executeMiddlewareBeforeRequest(request);
+      await this.executeMiddlewareBeforeRequest(request, response);
 
       // Is this request for the favicon?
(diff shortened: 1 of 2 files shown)


ebebbington:
Once this PR ios approved (if it is), we would need to sort out all middlewares in our mdidleware repos, but it means that things like serve static and rate limit middleware can be developed/merged

crookse:
we're not going with serve static if we're going to strip out the file serving logic

ebebbington:
whilst that is true, https://github.com/drashland/deno-drash-middleware/pull/126 still depends on this functionality (unless you know another way around it)?

ebebbington:
closing as we're going to land this in v2, and not much point of landing it now

PR #526, opened 2021-08-02
fix: getBodyFile only returning the first file is input is multiple

### LINKED ISSUE #525: Only retrieves one file even if file input is set to multiple
Eg when `<input type="file" multiple>`

The problem is because we're returning the first item of files on `Request#getBodyFile`

The fix would be just returning the whole files array, but i dont think we should do this, because if an input is single, then a user should get that file object and not an array containing it. My thinking is we make a getBodyFiles method, that holds the same logic as getBodyFile, and getBodyFile will just call getBodyFiles, and return the first item

ebebbington:
fixes #525

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -119,9 +119,22 @@ export class Request extends ServerRequest {
    * @return The file requested or `undefined` if not available.
    */
   public getBodyFile(input: string): FormFile | undefined {
+    const files = this.getBodyFiles(input);
+    return files ? files[0] : undefined;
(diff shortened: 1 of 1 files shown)


crookse (approved):
i'm guessing we need documentation for this new method. LGTM tho

ebebbington:
😤 good point tho, I’ll leave this pr open tho until the docs pr is merged, just as a reminder

crookse (approved):

PR #533, opened 2021-08-13
chore: hella stuff removed; made the server run again

crookse:
## Summary

* Took out all factory-related stuff
* Made the server run again


src/handlers/resource_handler.ts
```diff
@@ -186,11 +187,11 @@ export class ResourceHandler {
...
-    const clone = Drash.Prototype.clone(resource, {
-      path_parameters: pathParams,
-      request: request,
-      server: resource.server,
-    });
+    const clone = Drash.Prototype.clone(resource);
+
+    clone.path_parameters = pathParams;
+    clone.request = request;
+    clone.server = resource.server;
```
ebebbington, line 194: Not sure *why* this was changed, but it's not a problem, im sure you had your reasons and im not against it, just trynna understand

src/http/response.ts
```diff
@@ -1,28 +1,17 @@
...
- * This class represents a response that is generated by a resource -- to be
- * sent back to the client.
- */
-export class DrashResponse extends Response {
-  public headers = new Headers();
-  public status = 200;
-  #options: Drash.Interfaces.IResponseOptions = {};
+  public body: BodyInit | null | undefined;
+  public headers: Headers = new Headers();
+  public status: number = 200;
```
ebebbington, line 7: no need for this as its already present on the `Response` right? plus, it means we can parse `NativeResponse#body`

ebebbington (approved):
As i'm going to sleep, feel free to merge this PR anyways

src/handlers/resource_handler.ts
crookse: i didnt want the prototype to add the options automagically. we could probably just change this to `Object.create()` in the future

src/http/response.ts
crookse: we need this or else we can't use `this.response.body` in the resource. if we set `Resource.response = new Response()`, we can't mutate the `this.response.body` because it's readonly

PR #543, opened 2021-10-17
feat: Add redirect method

### LINKED ISSUE #522: Redirects
## Summary

What:

Add in `this.redirect()` method on the RESOURCE. In v1, it used to be on the response, but it doesn't make sense for the response object to have that method because the response just represents a response. The resource should be in charge of redirecting requests.

Why:

I removed the redirect logic. It needs to be implemented again.

## Acceptance Criteria

- [ ] Write documentation
- [ ] Write unit tests
- [ ] Write integration tests
- [ ] The resource handles redirects.

```typescript

class MyResource extends Drash.Resource {
  public GET() {
    return this.redirect("/some-path");
  }
}

```

ebebbington:
fixes #522

### DIFF
--- a/src/http/resource.ts
+++ b/src/http/resource.ts
@@ -11,4 +11,21 @@ import * as Drash from "../../mod.ts";
 export class Resource implements Drash.Interfaces.IResource {
   public services: Drash.Interfaces.IResourceServices = {};
   public paths: string[] = [];
+
+  /**
+   * Redirect the incoming request to another resource
+   *
+   * @example
+   * ```js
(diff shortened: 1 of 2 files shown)


crookse:
should we put this on the response object like before?

ebebbington:
> should we put this on the response object like before?

i went by:

> Add in this.redirect() method on the RESOURCE. In v1, it used to be on the response, but it doesn't make sense for the response object to have that method because the response just represents a response. The resource should be in charge of redirecting requests.

lol but it's definitely cleaner if on the response

PR #553, opened 2021-10-27
fix: Tengine not working; Remove render config

ebebbington:
fixes the fact that tengine doesn't run, because:

- it sets Jae's render method to the response object AFTER the resource is called, so in a resource when we call `response.render()`, it calls the responses method, which just returns false


crookse (dismissed):
LGTM. i'll do the docs

PR #568, opened 2021-11-08
feat: Add etag service

### LINKED ISSUE #559: ETag middleware
Original issue:  https://github.com/drashland/deno-drash/issues/17

Ported over here because @crookse decided to make it a middleware instead (which is a good call), because it means it's usuable by users, but doesn't affect drash's performance at all - win win scenario

**What**

Add a new middleware for supporting the ETag header

**Why**

* Improves caching = improves request speed
* Will not affect Drash's performance because Drash wouldn't have it built into the source code

**How**

* File structure would look like:
```
deno-drash-middleware/
    etag/ <-- etag code here
        tests/
        deps.ts
        some_related_file.ts
        another_related_file.ts
        mod.ts
        README.md
    ...
```
* Users would import the middleware like:

`import { ETag } from "https://deno.land/x/drash-middleware@v{version}/etag/mod.ts"`

* Place third-party modules in your middleware directory's deps.ts file (if used)

* Any Deno std module that's required by your package can be exported from the repo's root deps.ts file so that all middleware can import std modules from that single file.

**Acceptance Criteria**

- [ ] README.md file showing how to use the middleware. Drash is a documentation-driven codebase, so the documentation is very important.

- [ ] Develop ETag functionality

- [ ] Tests

ebebbington:
fixes #559

### DIFF
--- a/src/services/etag/etag.ts
+++ b/src/services/etag/etag.ts
@@ -20,16 +20,16 @@ export class EtagService extends Service implements IService {
       // when it's empty, we want to set a default etag
 
       // but if etag is already present on request, send a 304
-      const header = '"0-2jmj7l5rSw0yVb/vlWAYkK/YBwk"'
+      const header = '"0-2jmj7l5rSw0yVb/vlWAYkK/YBwk"';
       if (request.headers.get("if-none-match")) {
         response.status = 304;
         response.body = null;
-        const existingModifiedDate = this.#etags.get(header) as string // it will always be set due to this conditional
-        response.headers.set('last-modified', existingModifiedDate)
+        const existingModifiedDate = this.#etags.get(header) as string; // it will always be set due to this conditional
+        response.headers.set("last-modified", existingModifiedDate);
       } else { // set the NEW default etag
-        const date = new Date().toUTCString()
+        const date = new Date().toUTCString();
         response.headers.set("last-modified", date);
-        this.#etags.set(header, date)
+        this.#etags.set(header, date);
       }
       response.headers.set("etag", header);
       return;
@@ -62,22 +62,25 @@ export class EtagService extends Service implements IService {
         // no need to send body, send not modified
         response.status = 304;
         response.body = null;
-        response.headers.set('last-modified', this.#etags.get(header) as string)
+        response.headers.set(
+          "last-modified",
+          this.#etags.get(header) as string,
+        );
         return;
       } else {
         // res body is new
         response.status = 200;
-        const date = new Date().toUTCString()
-        this.#etags.set(header, date)
-        response.headers.set('last-modified', date)
+        const date = new Date().toUTCString();
+        this.#etags.set(header, date);
+        response.headers.set("last-modified", date);
         return;
       }
     }
 
     // else request doesnt have a new one so generate everything from scratch
     response.status = 200;
-    const date = new Date().toUTCString()
-    this.#etags.set(header, date)
+    const date = new Date().toUTCString();
+    this.#etags.set(header, date);
     response.headers.set("last-modified", date);
   }
 }
--- a/src/services/etag/etag.ts
+++ b/src/services/etag/etag.ts
@@ -4,35 +4,61 @@ import { createHash } from "./deps.ts";
 export class EtagService extends Service implements IService {
   #options: { weak: boolean };
 
-  #lastEtag = "";
-
   constructor(options: { weak: boolean } = { weak: false }) {
     super();
     this.#options = options;
   }
 
-  runAfterResource(_request: Request, response: Response) {
+  runAfterResource(request: Request, response: Response) {
+    // if response body is empty, send a default etag
     if (
       response.body === null ||
       (typeof response.body === "string" && response.body.length === 0)
     ) {
+      // when it's empty, we want to set a default etag
+
+      // but if etag is already present on request, send a 304
+      if (request.headers.get('if-none-match')) {
+        response.status = 304
+        response.body = null
+      } else { // set the NEW default etag
+        response.headers.set("Last-Modified", new Date().toUTCString());
+      }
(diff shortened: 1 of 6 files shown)


Guergeiro:
@ebebbington do you have the source for this implementation?

ebebbington:
> @ebebbington do you have the source for this implementation?

https://datatracker.ietf.org/doc/html/rfc7232 and https://github.com/jshttp/etag, which reminds me, i forgot to add a weak option

Guergeiro:
Don't know if this is the full implementation of the service. Maybe I'm reading it wrong, but we need to cache a resource `after` it's handled and, upon next request to the same resource, we should reuse if the headers allows it and the matching (weak or not) is true.

We should also send a `Last-Modified` header.

diveDylan:
@ebebbington  how about send response with `304` status code  when we resue the resource. For me , compare with the previous `etag`,  if the same add `304` status code for http cache notify

crookse:
seems like we need a caching layer to fully support this. am i correct in that assumption? @ebebbington @Guergeiro @diveDylan. i dont mean that drash would have core logic to support caching, but this service would need to add that in.

im wondering this because @diveDylan says we need to compare previous `etag` with current `etag` and that's the impression i was under when i first read about `etag`. then we need the last modified header stuff. would the service keep track of that info in a cache as well? or SHOULD the service be in charge of that?

ebebbington:
@Guergeiro the browser caches it itself if the etag header is set, but yeah a few bits to add, thanks for letting me know about that header you mentioned

Guergeiro:
> @Guergeiro the browser caches it itself if the etag header is set, but yeah a few bits to add, thanks for letting me know about that header you mentioned

That's not the issue. Shouldn't we save the etag generated? I know the browser caches the result, but what does it solve?

Example:
```
Iteration 1:

Browser request to /home ---> Drash Processes it ---> Drash Etags the Body ---> Drash returns body with etag.

Iteration 2:
Browser request to /home ---> Drash checks Etag ---> Drash returns 304.
```

ebebbington:
@Guergeiro no clue how it works to be honest, for us to know if the body has changed, we'd still have to allow the resource to run so we can check the body

Guergeiro:
@ebebbington @crookse found the reason why [here](https://stackoverflow.com/a/67929691)! We always learn.

Basically, I was over engineering stuff. What is missing:

- [ ] Making sure the headers implementation is right and perfect.
- [ ] Checking for body/headers same etag and returning 304 (without body) if true.

It's a nice read by the way, maybe something we link somewhere for future reference.

ebebbington:
@Guergeiro where are we checking the etag header? it isn't present on the inc request

diveDylan:
> @Guergeiro where are we checking the etag header? it isn't present on the inc request

@ebebbington 304 http code [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304),
```text
If-None-Match(previous etag from request header)
If-Modified-Since(previous Last-Modified from request header)

## only etag
getRequestHeader => If-None-Match === Etag => 304

## only Last-Modified
getRequestHeader => If-Modified-Since  ===Last-Modified => 304

## Last-Modified & Etag
getRequestHeader => ? => 304
```

ebebbington:
thanks @diveDylan :)

ebebbington:
fyi tests fail due to something in deno 1.16.1 or 1.16.0

ebebbington:
fyi @Guergeiro @diveDylan good for a final review if you wanna take a look

diveDylan:
> fyi @Guergeiro @diveDylan good for a final review if you wanna take a look

Http header field syntax is case-insensitive field,  [here](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2).
If drash's http header get is originnal, maybe you should lowercase the field name of the header, such as `If-No-Match` update to `if-no-match`

ebebbington:
@diveDylan thats a browser thing (and a Deno thing), doesn’t matter if you set it as case sensitive or not

say I set it as lower case, and in my browser I get the `If-None-Match` header, it’ll return the value I set for when I did `if-none-match`

src/services/etag/etag.ts
```diff
@@ -0,0 +1,66 @@
...
+      (typeof response.body === "string" && response.body.length === 0)
+    ) {
+      // when it's empty, we want to set a default etag
+
+      // but if etag is already present on request, send a 304
+      if (request.headers.get("if-none-match")) {
+        response.status = 304;
+        response.body = null;
+      } else { // set the NEW default etag
+        response.headers.set("Last-Modified", new Date().toUTCString());
```
Guergeiro, line 25: I would prefer this to be lower case.

src/services/etag/etag.ts
```diff
@@ -0,0 +1,66 @@
...
+        response.body = null;
+        response.headers.set("etag", header);
+        return;
+      }
+    }
+
+    // else it isn't the same, so set a NEW etag on res
+    response.status = 200;
+    response.headers.set("etag", header);
+    response.headers.set("Last-Modified", new Date().toUTCString());
```
Guergeiro, line 64: I would prefer this to be lower case.

Guergeiro (changes requested):
As you said @ebebbington it's irrelevant, but at least on our code we could keep somewhat of a consistent pattern.

src/services/etag/etag.ts
ebebbington: https://github.com/drashland/drash/pull/568/commits/10fb43cee7e0499411f7813158354ff1953876fb + below thread

Guergeiro (approved):
Seems okay to me!

crookse:
LGTM. did a `deno fmt`. i tested this in chrome and firefox. chrome always returns 200. firefox returns 304 as expected.

PR #572, opened 2021-11-13
fix: request content-length and content-type checks

crookse:
* Add check for `Content-Length` header
  * If `Content-Length` is null, then there is no body, so return `undefined` for the `#parsed_body`
  * If `Content-Length` is 0, then there is no body, so return `undefined` for the `#parsed_body`
* Add check for `Content-Type` header
  * If `Content-Length` is greater than 0, then there is a body and we must rely on the `Content-Type` header to parse the body correctly. If it is not present, then we throw a `400 Bad Request` error response.
* Fixed `request_test.ts` tests to include `Content-Length` so that they don't error out. The approach is hacky, but is only used in tests. In the real world, the `Content-Length` header should be provided by the client making the request.

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -147,8 +147,8 @@ export class DrashRequest extends Request {
    * @returns A parsed body based on the content type of the request body.
    */
   async #parseBody(): Promise<ParsedBody> {
-    const contentLength = this.headers.get("Content-Length")
-      ?? this.headers.get("content-length");
+    const contentLength = this.headers.get("Content-Length") ??
+      this.headers.get("content-length");
 
     // The Content-Length header indicates that the client is sending a body.
     // Some clients send a Content-Length header of "0", which indicates that
@@ -163,13 +163,13 @@ export class DrashRequest extends Request {
     // body, we need to know the Content-Type of the body. Otherwise, we have
     // no way of knowing how to parse it. We can assume, but let's just tell the
     // client to modify their request to include the Content-Type header.
-    const contentType = this.headers.get("Content-Type")
-      ?? this.headers.get("content-type");
+    const contentType = this.headers.get("Content-Type") ??
+      this.headers.get("content-type");
 
     if (!contentType) {
       throw new Errors.HttpError(
         400,
-        "Bad Request. The request body cannot be parsed due to the Content-Type header missing."
+        "Bad Request. The request body cannot be parsed due to the Content-Type header missing.",
       );
     }
 
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -147,13 +147,33 @@ export class DrashRequest extends Request {
    * @returns A parsed body based on the content type of the request body.
    */
   async #parseBody(): Promise<ParsedBody> {
-    try {
-      const contentType = this.headers.get(
-        "Content-Type",
+    const contentLength = this.headers.get("Content-Length")
+      ?? this.headers.get("content-length");
+
+    // The Content-Length header indicates that the client is sending a body.
+    // Some clients send a Content-Length header of "0", which indicates that
+    // there is in fact no body present with the request. We need to check for
+    // this or else we try to parse a body for no reason.
+    if (!contentLength || contentLength === "0") {
+      return undefined;
+    }
+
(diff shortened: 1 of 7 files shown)


src/http/request.ts
```diff
@@ -147,13 +147,33 @@ export class DrashRequest extends Request {
    * @returns A parsed body based on the content type of the request body.
    */
   async #parseBody(): Promise<ParsedBody> {
-    try {
-      const contentType = this.headers.get(
-        "Content-Type",
```
ebebbington, line 152: No need for checks on lowercase, it returns the same value for Content-Length and content-length (feel free to test it with a quick fetch in the browser or Deno)

src/http/request.ts
```diff
@@ -147,13 +147,33 @@ export class DrashRequest extends Request {
...
+    if (!contentLength || contentLength === "0") {
+      return undefined;
+    }
+
+    // If we get to this point, then that means there is a body (since the
+    // Content-Length header is greater than 0). In order for us to parse the
+    // body, we need to know the Content-Type of the body. Otherwise, we have
+    // no way of knowing how to parse it. We can assume, but let's just tell the
+    // client to modify their request to include the Content-Type header.
+    const contentType = this.headers.get("Content-Type") ??
```
ebebbington, line 166: See above

tests/integration/graphql_test.ts
```diff
@@ -1,5 +1,4 @@
 import { Rhum } from "../deps.ts";
```
ebebbington, line 1: Wait wasn’t this used to create the quick schema in the test?

tests/test_helpers.ts
```diff
@@ -89,35 +89,42 @@ export const makeRequest = {
...
+    if (options.body) {
+      options.body = JSON.stringify(options.body);
+    }
     return fetch(url, options);
   },
   delete(url: string, options: IMakeRequestOptions = {}) {
     options = Object.assign(options, {
       method: "DELETE",
     });
-    options.body = JSON.stringify(options.body);
```
ebebbington, line 113: Delete can’t have bodies right?

tests/test_helpers.ts
```diff
@@ -89,35 +89,42 @@ export const makeRequest = {
...
     return fetch(url, options);
   },
   patch(url: string, options: IMakeRequestOptions = {}) {
     options = Object.assign(options, {
       method: "PATCH",
     });
-    options.body = JSON.stringify(options.body);
+    if (options.body) {
+      options.body = JSON.stringify(options.body);
+    }
```
ebebbington, line 127: See above

ebebbington, line 127: See above

tests/integration/graphql_test.ts
crookse: this is now `GraphQL.buildSchema`. instead of just importing parts of the GraphQL library, i'm importing all of it and exporting it from `deps.ts`. this way, users can use all of GraphQL instead of only the things we import/export

tests/test_helpers.ts
crookse: they could. https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

tests/test_helpers.ts
crookse: patch can have a request body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH

crookse: im assuming this comment is meant for the patch request. if so, they can have a request body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH

src/http/request.ts
crookse: fixed: 87d964e. i opted to go with checking lowercase instead of uppercase

src/http/request.ts
crookse: 87d964e

tests/test_helpers.ts
ebebbington: woah, wack

ebebbington (approved):
noiceee :)

PR #577, opened 2021-11-13
chore(service/rate_limiter): split classes into separate files; use s…

crookse:
* Using snake case for properties
* Put MemoryStore into separate file and imported in into RateLimiterService
* Fixed typos
* Add file markers
* Organized code into public / private sections
* Use `#` instead of `private` access modifier

Tests are failing because of that content-length checking stuff which is addressed in #572


ebebbington (approved):
no biggie, address it if you want, regardless, happy with it all :)

PR #578, opened 2021-11-14
refactor: tengine

crookse:
i tried getting eta to work and it didn't work. also, they're using deno std 0.97.0. i tried getting dejs to work, but its render method returns a `StringReader` object. i think it's going to take more work to integrate other template engines with tengine and i think the implementation would be wonky af. if users want to use another template engine, then they can write their own service to use the template engine

so now we only support jae. might be worth just merging the jae code into the tengine file. we can save that for later though

docs were updated: https://github.com/drashland/website-v2/pull/40

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -130,7 +130,7 @@ export class DrashResponse {
    *
    *   https://github.com/drashland/deno-drash-middleware/tree/master/tengine
    */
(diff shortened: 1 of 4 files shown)


ebebbington (approved):
Damnnn alright fair enough

Issue #591, opened 2021-12-30
ErrorService to handle HttpErrors / introduce custom error handling


routmoute:
## Summary

What:

Possiblity to modify Errors messages:
For exemple, change
```
Error: Method Not Allowed
    at Server.<anonymous> (https://deno.land/x/drash@v2.3.0/src/http/server.ts:238:17)
    at async Server.#respond (https://deno.land/std@0.119.0/http/server.ts:298:18)
```
by
```json
{
    "error": "Method Not Allowed"
}
```
or what we want

Why:

Do not see details in production

## Acceptance Criteria

Below is a list of tasks that must be completed before this issue can be closed.

- [x] Write documentation
- [x] Write integration tests
- [x] Write feature

PR #592, opened 2021-12-30
feat: make error handling configurable

routmoute:
add Simple ErrorService possibility

See Issue #591


crookse:
hey @routmoute ! i like the idea behind this, but the current implementation is part of Drash's "core" code as opposed to being a separate service that can be plugged in through the `services` config. what i mean by that is:

### This Pull Request

```typescript
const server = new Server({
  protocol: "http",
  hostname: "localhost",
  port: 3000,
  resources: [],
  error_service: new MyErrorService() // <--- should be in services config
});
```

### Expected Implementation

```typescript
  protocol: "http",
  hostname: "localhost",
  port: 3000,
  resources: [],
  services: [ new MyErrorService() ] // <--- see here
});
```

reason we don't introduce much code into Drash's core is because we don't know for sure that developers want "built-in" features. the way Drash is currently built is so that developers are free to create their own services as well as a custom error service that meets their needs. however, we do feel that Drash should have an error handling layer that can be used easily, which is what you're introducing in this pull request. also, thank you for writing the documentation! much appreciated :).

so, all this being said, could you refactor your error service so that it can be plugged into the `services` config? if you run into any issues, feel free to message back and we'll respond asap. another option is we can merge your pull request in a feature branch, refactor some code, make a pull request to that feature branch, and have you review it. either option is fine with us, but we'd love your thoughts. let us know! thanks again!

routmoute:
I think I see, don't create another class but add the runOnError function directly on the services. I will refactor that, it will be better, thank you.

crookse:
hey @routmoute, so i had a talk with the core team and basically we think drash should have a built-in error handler and not an "error service". so your approach was mostly correct. sorry for misguiding you. the end goal would be a Drash.ErrorHandler class or something similar. that class would be instantiated and executed in server.ts' getHandler() method inside the catch block.

routmoute:
Arf, but, I think the second approach is better because developers can work errors differently if they are thrown on a particular method, path and then on the server.
On the first approach, we can only work at the server level.

crookse:
true. maybe we stick with the error service for now and see how that goes and if it needs to be changed we can always do that later and make another release

routmoute:
Actualy, I've just add a function `runOnError()` on Drash.Service class, this function is called in server.ts getHandler() catch block.

crookse:
hey @routmoute , so i pulled down your branch locally and tested it -- keeping developer user experience and maintainability in mind -- and during my testing, i realized using an "error service" will get us back to what we had in v1 (and we don't want to go back to what we had in v1).

in v1, the error handling got really messy and was hard to maintain because we had to run middleware during the error handling process. every time we came back to the code, we forgot how it worked and there were edge cases with it that we kept finding. in your branch, you basically have what we had in v1 (https://github.com/routmoute/drash/blob/errorService/src/http/server.ts#L354-L385). your code isn't wrong. it's exactly what's needed to have an "error service" in drash, but it's also something that we removed when going from v1 to v2.

@Guergeiro suggested that there be an error handling layer in drash and that it be a built-in feature that can be extended by developers and i disagreed because i thought having an "error service" would be best suited for configurability. after testing your code locally, i feel that @Guergeiro is correct and we should not use an "error service," but instead have an error handling layer that can be extended and configured by developers (if they choose to do so).

crookse:
feel free to hop in discord to discuss btw. we're pretty active in it and respond faster there. here's the link in case you're interested: https://discord.gg/DmjhWJt5

routmoute:
I will retry something

routmoute:
I've created an ExceptionLayer class can be exetended and declared on server 'exception' var.

what do you think about that ?

src/http/server.ts
```diff
@@ -331,10 +332,21 @@ export class Server {
...
           e.code = 500;
         }
-        return new Response(e.stack, {
-          status: e.code,
-          headers: response.headers,
-        });
+
+        if (exception) {
+          try {
+            await exception.catch(e, originalRequest, response);
```
crookse, line 338: this isn't documented in the codebase, but we try to instantiate things during compile time so that

1. the user doesn't have to do it
2. saves the server from having to build a new object for every request

could you make it so that the exception doesn't have to be instantiated by the user? that's the only feedback i have. id like to get thoughts on the naming of the class from @ebebbington and @Guergeiro

as far as implementation, LGTM. thanks @routmoute !

Guergeiro: You mean the `new Drash.ExceptionLayer().catch(...)`?

crookse: yea, but we instantiate it once and reuse it just like we do with resources

src/http/server.ts
```diff
@@ -331,10 +332,21 @@ export class Server {
...
+          } catch (error) {
+            if (isNaN(error.code)) {
```
Guergeiro, line 340: I know this is a leftover from previous code, but I detected it now.
I would rather use a class instance assertion than a property. It avoids collision.

```ts
if (isNaN(error.code) {
  // No good. The user can throw a custom error that has code as property.
  // I think a mongoose error would enter this if, with codes > 10000 ! Definitely not status codes.
}
```

```ts
if (error instanceof HttpError) {
  // We good. The class is in fact HttpError.
}
```

src/interfaces.ts
```diff
@@ -141,6 +141,7 @@ export interface IServerOptions {
   protocol: "http" | "https";
   resources: typeof Resource[];
   services?: Service[];
+  exception?: ExceptionLayer;
```
Guergeiro, line 144: This should be the `IExceptionLayer`. With that, I can create my own exception layer that does not extend the Drash layer, but rather completely overrides it.

A simple test with this would be a nice touch 😃

src/http/server.ts
Guergeiro: Would want a comment on that from @crookse or @ebebbington

Guergeiro (changes requested):

src/http/server.ts
Guergeiro: I agree with it. Let's avoid generating garbage for the garbage collector to collect 👍🏿

src/http/server.ts
crookse: i should've documented this, reason it's like this is because i ran into an issue where i just threw an Error without using Drash.Errors.HttpError. it had no code associated with it, so the catch block couldn't return a response with a status code. this is basically a "if an error was thrown and was not an HTTP error with an HTTP status code, then make it a 500 by default" block

src/http/server.ts
routmoute: ok, last commit instantiate ExceptionLayer on server.ts

src/interfaces.ts
routmoute: in last commit, I use typeof ExceptionLayer (like 'typeof Resource').

I added a test, I don't know if it's that you want.

src/http/server.ts
```diff
@@ -331,10 +332,21 @@ export class Server {
...
           e.code = 500;
         }
-        return new Response(e.stack, {
-          status: e.code,
-          headers: response.headers,
-        });
+
+        if (exception) {
+          try {
+            await (new exception()).catch(e, originalRequest, response);
```
ebebbington, line 338: feel like the `catch` method should be static?

ebebbington (commented):
Also in regards to https://github.com/drashland/drash/pull/592#discussion_r777264575, i think eric means, instantiate the exception class in the server constructor:

```ts
#exception: CustomException|null = null
constructor(...) {
  if (options.exception) {
    this.#exception = new options.exception()
  }
}
...

const exception = this.#exception
return function () {
  exception.catch(...)
}
```

But that wouldn't be a problem if the catch method was static anyways

src/http/server.ts
Guergeiro: > Also in regards to [#592 (comment)](https://github.com/drashland/drash/pull/592#discussion_r777264575), i think eric means, instantiate the exception class in the server constructor:
> 
> ```ts
> #exception: CustomException|null = null
> constructor(...) {
>   if (options.exception) {
>     this.#exception = new options.exception()
>   }
> }
> ...
> 
> const exception = this.#exception
> return function () {
>   exception.catch(...)
> }
> ```

Yup, this is what he means. Essentially the https://github.com/drashland/drash/pull/592/commits/8d4c75ea41671200f77735dc62f9feaeabf756fd was wrongly done.

> But that wouldn't be a problem if the catch method was static anyways

I would go against this. The purpose is that the users can extends the behaviour without breaking it. Static means the previous behaviour is lost.

src/http/server.ts
```diff
@@ -115,6 +120,9 @@ export class Server {
         patterns,
       });
     });
+    if (options.exception) {
+      this.#exception = new options.exception();
+    }
```
routmoute, line 125:
```typescript
if (options.exception) {
  this.#exception = new options.exception();
}
```
or
```typescript
if (this.#options.exception) {
  this.#exception = new this.#options.exception();
}
```
?
I hesitate ..

src/http/server.ts
routmoute: so what should I do ?

crookse: nothing for now. going to resolve this. this will be addressed in https://github.com/drashland/drash/issues/595

src/http/server.ts
crookse: this is good for now. resolving this. i say that because we check for `options.services` instead of `this.#options.services`.

src/interfaces.ts
crookse: @Guergeiro , i tested using `IExceptionLayer` as the expected type in the TS playground. seems to require `new MyExceptionClass()` to be used. see [playground here](https://www.typescriptlang.org/play?strict=false#code/JYOwLgpgTgZghgYwgAgJIFEAeSAOZgD2IAMnAJ7TIDeAUMsgnGAgBYAUEAXMgM5hSgA5gEpuANwLAAJgG4aAXxo0EAGzg8eyALJksufEVIUoyYAFscKiGYjhNGbBDyES5SrXqNm7Lr35DRZAlpajp6BiIeAisAOhUCQQ5hMMVFGlBIWEQUVAB5Z0jQ+ghHApBuB30XI2gFJVV1TQBlaDF3MIIyngr8gxAeOU9I-gBXBDACKDZOvu60XpceYSLwsBZgHhiZxeQAXmRtyMHkVPrh5DA95BAIAHdkFqg2qY9kEqqibh09Jz6aqAUwiAA).

src/http/server.ts
ebebbington: Resolving because the scalability of not making it static is better

src/interfaces.ts
Guergeiro: Yes. It's supposed to. Here's a [link](https://www.typescriptlang.org/play?strict=false#code/JYOwLgpgTgZghgYwgAgJIFEAeSAOZgD2IAMnAJ7TIDeAUMsgnGAgBYAUEAXMgM5hSgA5gEpuANwLAAJgG4aAXxo0EAGzg8eyALJksufEVIUoyYAFscKiGYjhNGbBDyES5SrXqNm7Lr35DRZAlpajp6BiIeAisAOhUCQQ5hMMVFZTUNbTIAYQBXPgIzPScDV2NkCExIEClNHWLnQzcTDwYmVg5uPgEQEVDw3lycaBivDohhOQGESOiIOIS2ACI4GEgoJeT6VKVQdfgkNAB5RpBNVsr9F24QCAB3ZDYY57goQR5uOBAyAG0AXWEyAAvAA+NANUpGaAKJSqdSaADK0DE7jCBFOH2OGKmETO-FyCDABCgbHRpUxqBO5MBrXoYBYwB4MTJLk0QOQLMiOJ2ykiYGQ-PZtweSKgKJJF0cp249SlkOaCmSvLxAoATMDkMLkKLxWxJVciDKcvkiUU5S4oVBFUA) with the same code, but the correct options signature.

crookse: ahh constructor. ok

src/http/server.ts
```diff
@@ -115,6 +120,10 @@ export class Server {
         patterns,
       });
     });
+
+    if (options.error_handler) {
+      this.#error_handler = new options.error_handler();
+    }
```
Guergeiro, line 126: No need for condition. Always apply the default if none exist.

```ts
this.#error_handler = new (options.error_handler() || Drash.ErrorHandler);
```

src/http/server.ts
```diff
@@ -173,6 +182,9 @@ export class Server {
   #getHandler(): (r: Request, connInfo: ConnInfo) => Promise<Response> {
     const resources = this.#resources;
     const serverServices = this.#options.services ?? [];
+    const errorHandler = this.#error_handler;
+    const defaultErrorHandler = new Drash.ErrorHandler();
```
Guergeiro, line 186: No need for this.

Guergeiro (changes requested):
In general, it's good for me. Logic wise.

This is some stuff I would prefer to simplify stuff. Basically, if no option for an ErrorHandler is passed, just automatically assign a default one. Then we can be assured it exists, otherwise it would fail in server setup.

src/http/server.ts
```diff
@@ -328,13 +340,17 @@ export class Server {
...
+          }
+        } else {
+          await defaultErrorHandler.catch(e, originalRequest, response);
         }
-        return new Response(e.stack, {
-          status: e.code,
-          headers: response.headers,
-        });
+
+        return new Response(response.body, response);
```
Guergeiro, line 353:
```ts
try {
  // stuff
} catch (e) {
  await this.#error_handler.catch(e, originalRequest, response);
}
```

Guergeiro (changes requested):

src/http/server.ts
Guergeiro: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgBIQDYYPbIN4BQyyAFpjgBQCUAXMgG7bAAmA3AQL4EEIZwDO-ZADFsuYAFsADhggSI4IeizZCxKQFcARhmAJS5bNXxFiyBNhD9ssgHQ4A5hQBEMMc6qmuXHn0HIAITgoZEkZOQUwJUMTdW1dfTIVYzUzCysbCHtsJ2ctYI8vTm4YDRAEMGBLZC0NYAxmaAoEAH46EAgAd2QKWz7gh346OBAATwBtAF0qZABeAD40QxnU9P4wBjnkDu7m5AAffcDgqnZielskylPiglr6xqhqdnuGptFsKiA

src/http/server.ts
routmoute: we use defaultErrorHandler when `this.#error_handler.catch()` or `errorHandler.catch()` (it's the same `const errorHandler = this.#error_handler;`) thrown an error

src/http/error_handler.ts
```diff
@@ -0,0 +1,21 @@
...
+   * not that `Drash.Request` object. Reason being, the `Drash.Request` object
+   * can also throw an error and would not be available as a parameter.
+   * @param response - The `Drash.Response` object.
+   */
+  public catch(
+    error: Errors.HttpError,
+    _request: Request,
+    response: Response,
+  ): void {
+    response.status = error.code ?? 500;
```
routmoute, line 18:
```typescript
if (error instanceof Errors.HttpError) {
  response.status = error.code;
} else {
  response.status = 500;
}
```

would be better or not ?
what happens if developer thrown a simple Error and not an HttpError ?

tests/integration/error_handler_test.ts
```diff
@@ -0,0 +1,127 @@
...
+import { ErrorHandler, Errors, Response, Server } from "../../mod.ts";
+
+class MyErrorHandler extends ErrorHandler {
+  public catch(error: Error, _request: Request, response: Response) {
+    if (error instanceof Errors.HttpError) {
+      response.status = error.code;
+    }
+    response.json({ error: error.message });
+  }
+}
```
routmoute, line 11: why error is Error here and is HttpError in default ErrorHandler ?
if is simple Error thrown, this ErrorHandler return 200 ?

src/http/error_handler.ts
routmoute: I think yes, to work correctly with simple Error

tests/integration/error_handler_test.ts
routmoute: I think use Error everywhere and set status to 500 if not an HttpError, see my last commits, it's better for me.

Guergeiro (approved):
I approve the architecture of this PR. Good job for me.

tests/integration/error_handler_test.ts
Guergeiro: So there's two approaches here:
1. Catch on every error. This means the server will never stop working, regardless of the error. If an error is thrown just send `500`.
2. Catch only HttpError. This can make the server crash if, for instance, a database errors and the developer doesn't worry about surrounding external calls with `try...catch`.

My opinion is that option 1 would be the most interesting. Ideally the server never goes down. If an error occurs that was not expected, and `Internal Server Error Exception` shall be sent to the user.

routmoute: Same opinion, that's exactly what I did.

crookse (approved):

crookse:
thank you @routmoute :)

Issue #599, opened 2022-01-25
`Response#file()` doesnt work on images


ebebbington:
Only tested on a jpg file, the code passes, no errors thrown, but the way we're reading **image** files is incorrect

The code needs to be `readFileSync`, not `readTextFileSync`

Tested myself and changing it still correctly displays js, css etc, but also will display pngs and jpgs

![image](https://user-images.githubusercontent.com/47337480/151033162-d784b30e-6b4a-4617-b166-fec829806a92.png)

Basic fix, but dont think we can properly test this, unless there's a way to know if a binary is readable? eg

```
test(() => {
  response.file('image.jpg')
  const body = response.body
  const isReadable = maybeSomeThirdPartyModule(body) // expected: true
})
```

crookse:
interesting. we could add a try catch to gracefully handle the error in case the file isn't readable, but should we?

ebebbington:
> interesting. we could add a try catch to gracefully handle the error in case the file isn't readable, but should we?

Would make sense but I think our beautiful lifecycle would already catch it and throw a http error

ebebbington:
The only test needed is that file() produces a binary, which it does now

If images/files are invalid, ie the content of the file, thats down to the user check and fix?

PR #602, opened 2022-01-31
fix: Response.file not working with images

ebebbington:
partially addresses #599 

Leaving issue open as we need to figur eout how to test this


crookse (approved):

crookse (commented):
Thought we shouldn't add a test since the error handler would handle this if the file isnt readable. that should be caught by the developer and would also have a proper error response thrown to the client

PR #606, opened 2022-02-09
fix: Files from bodyParam are corrupted

### LINKED ISSUE #604: feat: access raw request.body
## Summary

I would like to allow API consumers to upload a file directly in a request body, not via form encoding. In particular, I want to be able to process large file uploads as a stream. It seems that currently Drash always consumes `request.body`, so trying to access it myself results in the error `ReadableStream is locked.`

Why:

Some REST API methods make more sense as direct file uploads in the body, rather than attaching form-encoded files.

Some REST API methods should be able to handle huge (GB+) file uploads as streams, rather than storing file data in memory.

## Example Code

Most likely a flag would have to be set on the `Resource` or the method, to tell Drash not to process the body.

What I want to do:

```
const contentType = request.headers.get("Content-Type");
const bodyStream = request.body;
// Process the stream...
```

What I have to do for now as a workaround:

```
const formFile = request.bodyParam<{content: string, type: string, size: number}>('file');
const contentType = formFile.type;
const bodyStream = readableStreamFromIterable([ new TextEncoder().encode(formFile.content) ]);
// Process the "stream"...
```

ebebbington:
partially addresses #604 

````ts
const file = request.bodyParam<Drash.Types.BodyFile>('profile-picture');
if (!file) {
  //
}
console.log(file.content) // Uint8Array [ ... ]
```
Tested by writing the content to a file too, and works fine

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -1,18 +1,12 @@
-import { deferred, getCookies } from "../../deps.ts";
+import { getCookies } from "../../deps.ts";
 import { Errors } from "../../mod.ts";
 import type { ConnInfo } from "../../deps.ts";
+import { BodyFile } from "../types.ts";
 
 export type ParsedBody =
   | Record<string, string | BodyFile | BodyFile[]>
   | undefined
   | string;
-
-type BodyFile = {
-  content: unknown;
-  size: number;
-  type: string;
-  filename: string;
-};
 
 /**
  * A class that holds the representation of an incoming request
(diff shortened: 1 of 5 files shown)


ebebbington:
will write some docs tomorrow maybe

bradenmacdonald:
Thanks!

ebebbington:
docs: https://github.com/drashland/website-v2/pull/99

PR #612, opened 2022-02-22
Fix: Content-Disposition value for download method now used correctly

### LINKED ISSUE #611: Issue using download method downloads the file with `download` name
## Summary

Running this code https://drash.land/drash/v2.x/tutorials/responses/setting-the-body#download from the docs always generates a file named `download`.

## Steps To Reproduce The Bug

1. Copy and paste the code from the docs
2. Run the app
3. Open a browser tab with the application link
4. Will see all files downloaded as `download`

## Expected Behavior

It should use the filename from the function parameters

## Suggested Solution(s)

The issue is in the `Content-Disposition` header

## Screenshots

image

sant123:
Closes #611 

## Summary

According with https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_response_header_for_the_main_body the value for `Content-Disposition` should contain an equal sign and double quotes for the filename.

image

### DIFF
--- a/src/http/response.ts
+++ b/src/http/response.ts
@@ -55,7 +55,7 @@ export class DrashResponse {
   ): void {
     const filepathSplit = filepath.split("/");
     const filename = filepathSplit[filepathSplit.length - 1];
-    this.headers.set("Content-Disposition", `attachment; filename${filename}`);
+    this.headers.set("Content-Disposition", `attachment; filename="${filename}"`);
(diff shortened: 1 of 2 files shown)


ebebbington:
Oh nice thanks for the PR! Would you be able to add a unit test for this method, to ensure the header is correct?

ebebbington (approved):
Perfect :) Thanks again

PR #617, opened 2022-03-11
feat: Allow services to run at startup

crookse:
## Summary

* Placed types and interfaces to be reused in `interfaces.ts` and `types.ts`
* Added `runAtStartup` to service interface
* Added `Server.addResource()`.
* Refactored server to allow services to run at startup
* Removed `T` prefix from types. I don't know why I started doing this, but they're obviously types because they come from the `types.ts` file. woooooOoOopz.
* Started effort to move our server to expect a service interface as opposed to the `Service` class (kind of like how we're doing it for the error handler -- we expect an interface, not a class)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,13 +9,12 @@ doc_blocks_output_actual.json
 package-lock.json
 
 # directories
-.vim
+.idea
 .sass-cache
+.vim
+node_modules
 tmp*/*
 tmp/*
-node_modules
-.idea
-tmp-dir-for-testing-create-app
 
 # exceptions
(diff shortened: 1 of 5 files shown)


ebebbington (commented):
i think there will also be problems with `#getRequest`, because that method uses `this`, but inside deno's listen API, `this` will refer to the function as our server class is no longer in the scope

ebebbington (approved):
Docs needed for website too i assume?

crookse:
@ebebbington , yea docs are needed

PR #619, opened 2022-03-12
feat: Allow services to end lifecycle

### LINKED ISSUE #605: Allow services to send repsonses and end lifecycle
## Summary

Im sure we added something like this before, not sure what happened

What:

For example allow a service before resource to redirect, and not have the resource called

Why:

A real use case people have

ebebbington:
Closes #605

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ doc_blocks_output_actual.json
 package-lock.json
 
 # directories
+.vim
 .sass-cache
 tmp*/*
 tmp/*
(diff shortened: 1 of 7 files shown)


ebebbington:
```
server before
v
resource exists
v
class before
v
method exists
v
method before
v
call resource method
v
method after
v
class after
v
server after
```
If one of those steps fail, the link is broken and response is immediantly returned

crookse (commented):
LGTM. instead of marking a bunch of comments, i made a PR related to the comments i would've made: https://github.com/drashland/drash/pull/621

PR #621, opened 2022-03-19
refactor: add respond() function to consolidate returning responses in request handler

crookse:
## Summary

* Adds `respond(response: Drash.Response) => Response` in request handler so that services can do ...
  ```typescript
  if (endLifecycle) { return respond(response); }
  ```
  ... and the request handler and its `catch` block can do ...
  ```typescript
  return respond(response)
  ```
  The `respond(response: Drash.Response) => Response` also includes checking if responses were upgraded in case a service upgrades a response early in the lifecycle.
* Refactored `runServices()` to just return `end`. Reason being, currently in `service-send` branch, if a service throws an error, then we throw that error as soon as possible in the request handler. Instead, we can just not catch the error in `runServices()`, let `runServices()` throw that error to the request handler, then the request handler catches it in its `catch` block. This approach removes the following blocks for each `runServices()` call in the request handler:
  ```typescript
  if (servicesResponse.err) {
    throw servicesResponse.err;
  }
  ```
* Upgraded to Deno v1.20.1 and Deno Std 0.130.0
* Fixed CORS test because of Deno Std 0.130.0 automatically compression response bodies based on Accept-Encoding header. See https://deno.land/manual/runtime/http_server_apis#automatic-body-compression.

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -42,26 +42,29 @@ async function runServices(
   request: Drash.Request,
   response: Drash.Response,
   serviceMethod: "runBeforeResource" | "runAfterResource",
-): Promise<{ err: Error | null; end: boolean }> {
-  let err: Error | null = null;
+): Promise<boolean> {
   let end = false;
+
+  // There are two ways a service can short-circuit the
+  // request-resource-response lifecycle:
+  //
+  //   1. The service throws an error.
+  //   2. The service calls `this.end()`.
+  //
+  // If the service throws an error, then the request handler we pass in to `new
+  // StdServer()` will catch it and return a response.
+  //
+  // If the service calls `this.send()`, then the request handler we pass in to
+  // `new StdServer()` will return `new Response()`.
   for (const Service of Services) {
-    try {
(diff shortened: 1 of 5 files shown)


src/http/server.ts
```diff
@@ -42,26 +42,29 @@ async function runServices(
...
+  // request-resource-response lifecycle:
+  //
+  //   1. The service throws an error.
+  //   2. The service calls `this.end()`.
+  //
+  // If the service throws an error, then the request handler we pass in to `new
+  // StdServer()` will catch it and return a response.
+  //
+  // If the service calls `this.send()`, then the request handler we pass in to
+  // `new StdServer()` will return `new Response()`.
```
ebebbington, line 58: good shout, now we dont need to hold onto the error until other services are ran, it can be caught by our main lifecycle logic

ebebbington (approved):
you just improved my crap code 💩 nice job, really **DRY**

crookse:
> you just improved my crap code 💩 nice job, really **DRY**

:heart:

PR #625, opened 2022-04-03
fix: change service.end() to be request.end()

### LINKED ISSUE #624: Service.end() keeps the send property set to true
## Summary

When calling `someService.end()`, it sets the `send` property to `true` and it never gets reset. This means subsequent requests that go through `someService` will cause the Drash server to see `someService` is set to end the lifecycle early.

## Expected Behavior/Suggested Solution

Services are singletons, so their `send` property needs to be reset. Also, I didn't realize this, but we're setting state in a singleton, which could lead to more issues. We only have one property that determines the state of a service, so maybe it's good for now, but we should try to find a better way to know that a service wants to end the lifecycle without modifying its state.

crookse:
Fixes #624 

## Documentation

https://github.com/drashland/website-v2/pull/124

## Summary

I changed the API for ending a request lifecycle early. Instead of calling `this.end()` in a service, we call `request.end()` in the service. Reason being:

If one request comes in and the services ends early and there's another request that comes in and uses that service when it's `end_lifecycle` flag is still `true`, then that second request will end early too.

Each request needs to work in different contexts. Since services are singletons, each request will see the same service and its context. Any properties set in a service would be applicable to all requests, hence why I removed the property and moved it to the request class since each request is a siloed context.

## Test Findings

I added logs to see if a first request (takes 5 seconds and `this.end()` is called in the service) would cause a second request to see that the service had `this.end()` already called. The results were:

```
first request came in. calling this.end()
value of this.end_lifecycle: true
setting timeout of 5 seconds before ending request.
second request came in. <---------------------------------first request is still processing and set this.end_lifecycle to true
value of this.end_lifecycle: true <---------------------- this second request sees that this.end_lifecycle is still true
second request got a response
first request got a response
```

The fix results in the following:

```
first request came in. calling request.end()
value of request.end_lifecycle: true
setting timeout of 5 seconds before ending request.
second request came in. <-------------------------------- first request is still processing and set request.end_lifecycle to true
value of request.end_lifecycle: false <------------------ this second request sees that request.end_lifecycle is not true. this is the expected behavior.
second request got a response
first request got a response
```


ebebbington (approved):

Guergeiro (approved):

ebebbington (approved):
i think thats fine :) as long as it's simple for a user to do `this.endTheBloodyThing()` then we gucci

PR #634, opened 2022-04-22
fix: ensure consistent falsy return values for body params

### LINKED ISSUE #623: bodyParam returns undefined for null-value params
`bodyParam` returns `undefined` for param_names that actually set to `null` in application/json post-request
at the same time correct value can be obtained by calling `bodyAll()[param_name]`

e.g.
```ts
request.bodyAll()           // { foo: null }
request.bodyAll()['foo']    // null
request.bodyParam('foo')    // undefined
```

is the `bodyParam` function currently behaving correctly?
```ts
// src/http/request.ts#L132
return this.#parsed_body![name] as unknown as T ?? undefined;
```

ebebbington:
closes #623

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -12,7 +12,7 @@ export type ParsedBody =
  * A class that holds the representation of an incoming request
  */
 export class DrashRequest extends Request {
-  #parsed_body!: ParsedBody;
+  #parsed_body?: ParsedBody;
   readonly #path_params: Map<string, string>;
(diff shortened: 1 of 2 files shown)


crookse (approved):
LGTM. although i did work on it some. the tests do exercise that we return the following:

- `null` body param value returns `null`
- `false` body param value returns `false`
- `undefined` body param value returns `undefined`
- no value set in body param returns `undefined`

PR #637, opened 2022-05-14
perf: don't find resource if already matched to previous request URL

### LINKED ISSUE #636: perf: dont try to find resource if it was already found previously
## Summary

What:

We used to have tracking in place where if a request came in and was matched to a resource, then we would store the request URL in a map and its value would be the resource.

Why:

We should introduce this logic again for to improve performance.

## Notes

- There was a concern where if we associated a resource with a request URL, that the request would receive a cached response. This isn't an issue because we are only associating the resource with the request, not the response. That means when a request comes in and is matched to a resource, we associate that resource with that request's URL. If the resource does a database call and returns some dynamic data, that dynamic data would still be "good" because we are only associating the request URL to the resource and not associating responses to request URLs.

## Acceptance Criteria

- [ ] Write unit tests
- [ ] Write integration tests

crookse:
Closes #636 

Before:

```
Running 10s test @ http://localhost:1447?test=test
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   141.61ms  175.94ms   1.97s    95.86%
    Req/Sec   424.22     85.40   640.00     73.37%
  8498 requests in 10.10s, 4.51MB read
  Socket errors: connect 0, read 0, write 0, timeout 11
  Non-2xx or 3xx responses: 8498
Requests/sec:    841.39
Transfer/sec:    456.85K
```

After:

```
Running 10s test @ http://localhost:1447
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.46ms    2.91ms 100.13ms   99.54%
    Req/Sec    21.55k     1.24k   22.50k    98.02%
  433043 requests in 10.10s, 56.58MB read
Requests/sec:  42867.70
Transfer/sec:      5.60MB
```

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -76,6 +76,13 @@ export class Server {
   #default_error_handler = new Drash.ErrorHandler();
 
   /**
+   * Property to track request URLs to resources. This is used so that the
+   * server does not have to find a resource if it was already matched to a
+   * previous request's URL.
+   */
+  #request_to_resource_map = new Map<string, Drash.Interfaces.IResourceAndParams>();
+
+  /**
    * The internal and external services used by this server. Internal services
    * are ones created by Drash. External services are ones specified by the
    * user.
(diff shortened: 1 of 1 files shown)


Guergeiro:
What about [Moogle](https://github.com/drashland/moogle)? Won't it help in anyway?

crookse:
Moogle would help but the code needed to make this happen is tiny compared to what Moogle has.

Guergeiro:
And I agree with you. Just checking if it would be better or not.

Guergeiro (approved):

PR #638, opened 2022-05-14
fix: check for non-empty accept header before performing accept header flow

### LINKED ISSUE #635: bug: 406 thrown for empty string accept header
Was doing benchmarks with autocannon and wrk and received 406s. this was due to the accept header being "" which made the request handler think it had to go through the accept header flow -- returning 406s.

Need to check for empty string before running through the accept header flow.

crookse:
Closes #635

Before:

```
Running 5s test @ http://localhost:1447?test=test
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   162.52ms  253.89ms   1.99s    92.58%
    Req/Sec   473.68     72.50   643.00     73.00%
  4796 requests in 5.09s, 2.54MB read
  Non-2xx or 3xx responses: 4796
Requests/sec:    941.74
Transfer/sec:    511.34KB
```

After:

```
Running 5s test @ http://localhost:1447?test=test
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.74ms    5.18ms 124.77ms   99.08%
    Req/Sec    14.80k     1.48k   15.78k    90.20%
  150207 requests in 5.10s, 19.63MB read
Requests/sec:  29431.85
Transfer/sec:      3.85MB
```

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -371,15 +371,14 @@ export class Server {
         return this.#respond(response);
       }
 
-      const accept = request.headers.get("accept") ?? "";
-      const contentType = response.headers.get("content-type") ?? "";
-      if (accept && accept.trim() != "" && accept.includes("*/*") === false) {
-        if (accept.includes(contentType) === false) {
-          throw new Drash.Errors.HttpError(
-            406,
-            "The requested resource is only capable of returning content that is not acceptable according to the request's Accept headers.",
-          );
-        }
+      const requestAcceptHeader = request.headers.get("accept");
+      const responseContentTypeHeader = response.headers.get("content-type");
+
+      if (requestAcceptHeader && responseContentTypeHeader) {
+        this.#verifyAcceptHeader(
+          requestAcceptHeader,
+          responseContentTypeHeader,
+        );
       }
 
       return this.#respond(response);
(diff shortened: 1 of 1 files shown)


src/http/server.ts
```diff
@@ -373,7 +373,7 @@ export class Server {
 
       const accept = request.headers.get("accept") ?? "";
       const contentType = response.headers.get("content-type") ?? "";
-      if (accept.includes("*/*") === false) {
+      if (accept && accept.trim() != "" && accept.includes("*/*") === false) {
```
Guergeiro, line 376: God damn this conditions...

`accept` will always be defined because of line 374, therefore checking for it being defined is not required.

Also, `accept.trim()` should be compared with `!==` not `!=`.

Solutions for this condition may be:
```ts
// Solution 1
if (accept.trim() !== "" && accept.includes("*/*") === false) {
  // Stuff
}

// Solution 2
if (accept.trim().length !== 0 && accept.includes("*/*") === false) {
  // Stuff
}
```

Guergeiro (changes requested):

src/http/server.ts
crookse: OHHHHH wow. i didn't see the `?? ""` part. that def changes things

ebebbington:
Mad this changes the speed so much

src/http/server.ts
crookse: how's this @Guergeiro ? https://github.com/drashland/drash/pull/638/commits/95a419bdf5a60ecff892c9208a6092a27e8dd1d1

Guergeiro (approved):
Go go go!

PR #649, opened 2022-05-29
feat(http/request): allow access to the original request

### LINKED ISSUE #604: feat: access raw request.body

crookse:
Closes #604 

## TODO

- [x] Write documentation
  - PR: https://github.com/drashland/website-v2/pull/169
  - Page: https://drash-land-git-drash-v2-raw-request-access-drash-land.vercel.app/drash/v2.x/tutorials/requests/handling-the-original-request-body
- [x] Write unit tests

## Summary

* Makes `Drash.Request.original` property to store the original `Request`
* `Drash.Request.original` can be used to access the raw request body
* `Drash.Request.parseBody()` does not make `Drash.Request.original.bodyUsed` to be `true`, so it can be parsed manually by the user

## Other Notes

This will allow users to access the original if needed. For example, #604 requests access to the raw body. This can now be done like so:

```typescript
import * as Drash from "...";

class SomeResource extends Drash.Resource {
  public paths = ["/something"];

  public async POST(req: Drash.Request, res: Drash.Response): Promise<void> {
    // Access Drash's internal request helper method to get the body params
    console.log(req.bodyAll()); // Reads the Drash.Request body
    console.log(req.bodyUsed);  // true

    // Now access the original and parse the body manually
    console.log(req.original.bodyUsed);     // false --> The original request body is intact and has not been read yet
    // Now read the original request body
    console.log(await req.original.json()); // { hello: "world" }
    console.log(req.original.bodyUsed);     // true ---> Now the original request body has been read

    res.text("GET /something");
  }
}

  hostname: "localhost",
  port: 1447,
  protocol: "http",
  resources: [SomeResource],
});

server.run();

```

Example curl request for the above sample application:

```
$ curl -X POST -H 'Content-Type: application/json' localhost:1447/something --data '{"hello":"world"}'
```


Guergeiro (approved):
Go Go Go!

bradenmacdonald:
Nice, thanks!

PR #656, opened 2022-06-19
feat: Check for tailing/leading slashes in Jae

### LINKED ISSUE #648: feat: check for trailing/leading slash in jae
## Summary

What:

Jae (in Tengine) line 76 concatenates `this.views_path` and `template`. The expected strings should look like:

```
this.view_path = "./views";

template = "/some-template.html"
```

... so that the concatenation becomes `./views/some-template.html`. However, users most likely do:

```
this.view_path = "./views";

template = "some-template.html"
```

... so that the concatenation becomes `./viewssome-template.html`. This results in an error.

We need to make sure of the following:
  - If `this.views_path` has a trailing slash and `template` has a leading slash, then remove one of them
  - If `this.views_path` does not have a trailing slash and `template` does not have a leading slash, then add the slash.

This makes `Jae` less error prone and it does not matter if the user has a random combination of trailing/leading slashes -- `Jae` wlll figure out what to do.

Why:

Users expect to use Tengine differently.

## Acceptance Criteria

- [ ] Write documentation
- [ ] Write unit tests
- [ ] Write integration tests

ebebbington:
closes #648

### DIFF
--- a/src/services/tengine/jae.ts
+++ b/src/services/tengine/jae.ts
@@ -48,12 +48,12 @@ export class Jae {
     let cursor = 0;
     let match;
     let filepath = this.views_path;
-    if (this.views_path.endsWith('/') && template.startsWith('/')) {
-      filepath += template.slice(1)
-    } else if (!this.views_path.endsWith('/') && !template.startsWith('/')) {
-      filepath += `/${template}`
+    if (this.views_path.endsWith("/") && template.startsWith("/")) {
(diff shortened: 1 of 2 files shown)


crookse (approved):
LGTM ! 😭 this is going to be super helpful imo

ebebbington:
@Guergeiro windows tests pass :)

PR #665, opened 2022-07-02
Run runAfterResource for successful services, fix #663

### LINKED ISSUE #663: bug: ResponseTimeService not work when an exception occurs
## Summary

The Service not work when an exception occurs, such as 40X or 50X.

## Steps To Reproduce The Bug

```sh
$ curl -I 0.0.0.0:8000/404                                                                                                  
HTTP/1.1 200 OK
content-type: text/html
vary: Accept-Encoding
content-length: 569
date: Fri, 01 Jul 2022 15:10:59 GMT
```

## Expected Behavior

The Service should work for all status code.

## Suggested Solution(s)

When a service or resource behind a service throws an exception, 
it should also run the `runAfterResource` of this service after calling `ErrorHandler`.

Otherwise, middlewares such as Logger or ResponseTime  will not work when exception occurs.

biluohc:
Fixes #663

### DIFF
--- a/src/http/error_handler.ts
+++ b/src/http/error_handler.ts
@@ -1,6 +1,5 @@
-import { Errors, IErrorHandler, Response } from "../../mod.ts";
+import { Errors, IErrorHandler, Request, Response } from "../../mod.ts";
 import { STATUS_TEXT } from "../../deps.ts";
-import type { ConnInfo } from "../../deps.ts";
 
 export class ErrorHandler implements IErrorHandler {
   /**
@@ -16,7 +15,6 @@ export class ErrorHandler implements IErrorHandler {
     error: Error,
     _request: Request,
     response: Response,
-    _connInfo: ConnInfo,
   ): void {
     const errorMessage = error.stack ?? "Error: Unknown Error";
 
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -81,14 +81,22 @@ export class DrashRequest extends Request {
     connInfo: ConnInfo,
   ) {
     const req = new DrashRequest(request, pathParams, connInfo);
+    await req.prepare();
+    return req;
+  }
+
+  /**
+   * Try prepare ParsedBody for resources, etc
+   */
+  async prepare() {
     // This is here because `parseBody` is async. We can't parse the request
     // body on the fly as we dont want users to have to use await when getting a
     // body param.
+    const req = this.original;
     const contentLength = req.headers.get("content-length") ?? "0";
     if (req.body && req.bodyUsed === false && contentLength !== "0") {
-      req.#parsed_body = await req.#parseBody();
+      this.#parsed_body = await this.#parseBody();
     }
-    return req;
   }
 
   //////////////////////////////////////////////////////////////////////////////
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -2,10 +2,11 @@ import * as Drash from "../../mod.ts";
 import { ConnInfo, StdServer } from "../../deps.ts";
 
 async function runServices(
-  Services: Drash.Interfaces.IService[],
+  services: Drash.Interfaces.IService[],
+  serviceMethod: "runBeforeResource" | "runAfterResource",
   request: Drash.Request,
   response: Drash.Response,
-  serviceMethod: "runBeforeResource" | "runAfterResource",
+  afters: Drash.Interfaces.IService[],
 ): Promise<void> {
   // There are two ways a service can short-circuit the
   // request-resource-response lifecycle:
@@ -18,16 +19,25 @@ async function runServices(
   //
   // If the service calls `request.end()`, then the request handler we pass in
   // to `new StdServer()` will return `new Response()`.
-  for (const Service of Services) {
-    if (serviceMethod in Service) {
-      await Service[serviceMethod]!(request, response);
+  for (const service of services) {
+    if (serviceMethod in service) {
+      await service[serviceMethod]!(request, response);
+    }
+
+    if (serviceMethod === "runBeforeResource") {
       if (request.end_lifecycle) {
         break;
+      }
+
+      // should run runAfterResource for this service
+      if (
+        "runAfterResource" in service
+      ) {
+        afters.push(service);
       }
     }
   }
 }
-
 /**
  * This class handles the entire request-resource-response lifecycle. It is in
  * charge of handling incoming requests, matching them to resources for further
@@ -264,7 +274,7 @@ export class Server {
     connInfo: ConnInfo,
   ): Promise<Response> {
     // Grab resource and path params
-    const resourceAndParams = this.#getResourceAndParams(
+    const { resource, pathParams } = this.#getResourceAndParams(
       originalRequest.url,
       this.#resources,
     ) ?? {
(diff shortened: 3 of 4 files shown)


crookse:
hey @biluohc ! while i agree that this is an issue that should be fixed, i do think we should investigate a way to handle this without touching the core code (anything in the `src` directory that is not in the `services` directory is core code). if the core code has to be modified, then i think modifying it just a tiny bit should suffice. for context, the reason we stopped drash from running services after an error is thrown is because of the "fail fast" model we agreed to implement. however, we overlooked the `ResponseTimeService` and the `DexterService` when doing this. those seem to be valid use cases that drash should run even when an error is thrown.

@Guergeiro @ebebbington , instead of changing core code, i propose we modify `ResponseTimeService` to allow the `start_time` property (it's actually `startTime` in the current implementation) to be accessed. it would look like:

<details>
<summary>Proposed new ResponseTimeService with access to start_time</summary>

```typescript
import * as Drash from "https://deno.land/x/drash@v2.6.0/mod.ts";

// Proposed new ResponseTimeService with access to the `start_time` property
export class ResponseTimeService extends Drash.Service {
  #end_time = 0;
  #start_time = 0;

  get start_time() {
    return this.#start_time;
  }

  public runBeforeResource(): void {
    this.#start_time = new Date().getTime();
  }

  public runAfterResource(
    _request: Drash.Request,
    response: Drash.Response
  ): void {
    this.#end_time = new Date().getTime();
    const time = (this.#end_time - this.#start_time) + "ms";
    response.headers.set("X-Response-Time", time.toString());
  }
}

// Instantiate the proposed new ResponseTimeService
const responseTimeService = new ResponseTimeService();

// Provide an error handler to catch errors and finish adding the
// X-Response-Time header
class MyErrorHandler extends Drash.ErrorHandler {
  public catch(error: Error, request: Request, response: Drash.Response): void {
    this.#setResponseTime(response);

    response.json({
      error: error.message,
    });
  }

  /**
   * Set the X-Response-Time header in case an error occurs during the
   * request-resource-response lifecycle.
   *
   * @param response - The response to add the X-Response-Time header.
   */
  #setResponseTime(response: Drash.Response): void {
    const endTime = new Date().getTime();

    // Access the `ResponseTimeService.start_time` to complete the
    // X-Response-Time header value
    const responseTime = `${(endTime - responseTimeService.start_time)}ms`;

    response.headers.set("X-Response-Time", responseTime);
  }
}

// Add a resource to exercise the code. This resource has a timeout so that the
// X-Response-Time header has more than 0ms recorded.
class HomeResource extends Drash.Resource {
  public paths = ["/"];

  public async GET(request: Drash.Request, response: Drash.Response): Promise<void> {
    const p = new Promise((resolve) => {
      setTimeout(() => {
        resolve(true);
      }, 250);
    });

    // Wait 250ms before throwing the below error
    await p;

    throw new Error("Boom!");
  }
}

// Start and run the server

  error_handler: MyErrorHandler,
  hostname: "0.0.0.0",
  port: 1447,
  protocol: "http",
  resources: [
    HomeResource,
  ],
  services: [
    responseTimeService,
  ],
});

server.run();

```
</details>

this above approach could possibly be used for `DexterService` as well. however, i think this degrades developer UX because we leave it up to the developer to run the services in an `ErrorHandler` implementation in case an error occurs.

the "modify core" option off the top of my head looks something like this:

<details>
<summary>Modifying server.ts#handleRequest() catch block</summary>

```diff
// server.ts#handleRequest()

    } catch (e) {
      try {
+       for (const service of this.#services) {
+         if ("runAfterError" in service) {
+           service.runAfterError(e, originalRequest, response, connInfo);
+         }
+       }

        await this.#error_handler.catch(e, originalRequest, response, connInfo);
      } catch (e) {
+       for (const service of this.#services) {
+         if ("runAfterError" in service) {
+           service.runAfterError(e, originalRequest, response, connInfo);
+         }
+       }

        await this.#default_error_handler.catch(
          e,
          originalRequest,
          response,
          connInfo,
        );
      }

      return this.#respond(response);
    }
  }
```
</details>

thoughts? @biluohc , please feel free to give your input as well since you raised this issue.

biluohc:
@crookse  

## About call services in error_handler
It is really not feasible for developers to call services in `error_handler`. 
First, there is no context (it is strange to use the instance of `responsetime` directly).
Second, there is a lack of composability (for example, users need to try catch exceptions,  multiple services need to be handled in this way  and some services depends on other services, the meaning of the framework is actually gone))

## About `runAfterError`
The user can see the status code of the response in the runAfterResource, so the user can decide how to deal with it. In my opinion, it is not very necessary to add a `runAfterError`, but I also accept it.

## About `ErrorHandler`
All `Service` and `ErrorHandler` use the `Drash.Resquest` is more consistent, which is helpful for users to understand and use. The change is compatible with the current interface.

crookse:
@biluohc , i do agree the first option (calling services in the error handler) degrades developer UX. my goal with that first option was to propose a way that does not modify core and simply give an idea of how we would do it.

for adding a `runAfterError`, i'm not proposing that we add that. im just giving an idea of what it could like look and what it looked like to me off the top of my head -- just a starting point to spin up ideas and spark discussions on how to best handle running services when an error occurs. 

im not sure what you mean by the following. could you explain more please?:

> In your code, runAfterError can't get actually Response because users could modify Reponse in error_handler(so it not suitable for logger. The errorHandler of users maybe slow, so it is not suitable for responseTime running in front of the errorHandler).

when you mention the response time service not being suitable to be in front of the error handler, i get what you're saying, but the response time can mean many things to many developers. for example, it could mean the following:
- when the request enters a reverse proxy, goes to the backend services, comes back to the reverse proxy, and leaves the reverse proxy
- when the request enters an API gateway, goes to the backend services, comes back to the API gateway, and leaves the API gateway
- when the request enters the framework and leaves the framework
- when the request enters the request handler (inside the framework) and leaves the request handler
- when the request enters the resource and leaves the resource

drash doesn't really give you the option to inject the response time service (this is true for all services) exactly where you want it to go if you want to (or don't want to) include drash's internals in your response time metrics. i think if you wanted to be granular like that with the response time service, the best option would be to create your own implementation of the `Drash.Server` class. i did check the documentation and we don't state what i just said, so i think that's an action item for us to do.

as for your questions:

1. Can other levels of service use runAfterError?
  - Answer: possibly. it would really depend on what the request-resource-response lifecycle flow looks like after refactoring. off the top of my head, i don't know the answer to this question.
2. Can runAfterError throw exceptions?
  - Answer: yes. we don't restrict developers from throwing errors anywhere in their services and it is left to the developers to handle those errors
3. Some services may not run runBeforeResource due to exceptions of front service, so should not run it's runAfterError?
  - Answer: it depends on uses cases. for example, if a ServiceA throws an error and ServiceLogger (writes request logs) comes after ServiceA, then i think ServiceLogger should still write request logs cause in that case it makes sense.

> All Service and ErrorHandler use the Drash.Resquest is more consistent, which is helpful for users to understand and use. The change is compatible with the current interface.

The `Drash.Request` cannot be used in the error handler because it is created in the `try-catch` block of the server's request handler. if it throws an error, then it's not available to the error handler. the reason [the creation of the request object](https://github.com/drashland/drash/blob/main/src/http/server.ts#L282-L286) is in the `try-catch` block is because it can throw errors when drash tries to read the body. for example, if there is a body but there is no `content-type` header, then drash has no idea how to parse the body so it throws an error (now the `Drash.Request` object is not available to the error handler).

biluohc:
#### About  runAfterError not suitable for logger
Suppose the logger needs to display the final status code, 
but user's  `errorHander` may modify the response, so httpError got in `runAfterError` may not be the final status code.

##### About ErrorHandler  use Drash.Request 

You can take a look at my code, it separate the parseBody from create, and its creation will no longer require async and try-catch.


https://github.com/drashland/drash/blob/aae1065a1da880d778397751dc83b066babb8b2a/src/http/request.ts#L78-L100

 ##### async #handleRequest()
```ts
    // Construct request and response objects to pass to services and resource
    // Keep response top level so we can reuse the headers should an error be thrown
    // in the try
    const response = new Drash.Response();
    const request = new Drash.Request(originalRequest, pathParams, connInfo);

    const afters: Drash.Interfaces.IService[] = [];
    try {
      // todo : move the prepare to befor calling resource, Compatibility needs to be considered
      // parseBody, etc
      await request.prepare();

      // Run server-level services (before we get to the resource)
      await runServices(..)
     } catch (e) {
     }
}
```

PR #666, opened 2022-07-03
refactor(http/server): allow server to be extended so that users can control handleRequest()

crookse:
## Summary

- Makes the necessary changes to allow extending `Drash.Server` with "little" effort. I say "little" because the only method that needs to be implemented is the `handleRequest()` method. Right now, if a user were to extend `Drash.Server`, they'd have to recreate a bunch of the internals like `runServices()`, which lives outside the `Server` class. In this PR, users can now extend `Drash.Server` and provide their own implementation of `handleRequest()` -- also being able to call the necessary methods to recreate the request-resource-response lifecycle (e.g., they can call `this.runServices()`). Reason for this is to allow people to write what they think is best for handling a request for their server's implementation while still taking advantage of Drash's `Request`, `Resource`, `Response`, and `Service` classes.
- This also makes a change to the method name `this.#getResourceAndParams()` to `this.getResourceProxy()`. Reason I chose proxy was because it seems more fitting, but open to suggestions.
- If we decide to go through with this, then we can't hide what we're currently exposing in this PR, so I think we need to scrutinize this PR before releasing it. Otherwise we might run into a "dang we gotta make a breaking change and go to v3" situation.
- Did some alphabetizing, so the diffs are bigger than they should be 🥶

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -1,33 +1,6 @@
 import * as Drash from "../../mod.ts";
 import { ConnInfo, StdServer } from "../../deps.ts";
 
-async function runServices(
-  Services: Drash.Interfaces.IService[],
-  request: Drash.Request,
-  response: Drash.Response,
-  serviceMethod: "runBeforeResource" | "runAfterResource",
-): Promise<void> {
-  // There are two ways a service can short-circuit the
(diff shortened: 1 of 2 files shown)


crookse:
nope.

PR #671, opened 2022-07-08
feat(http/server): add request optionals to control reading the body

### LINKED ISSUE #669: feat(http/server): add new config for control flow changes
## Summary

What:

#604 requested access to the request body. This was solved by allowing the original request to be cloned and accessed via `request.original`. While this kept the body intact and prevented `bodyUsed` from being `true`, `Drash.Request.parseBody()` still runs and expects the request body to be a specific set of content-types. This needs to be changed. However, this is needs to be a non-breaking change and means we might need to introduce a server config to change the request's flow. For example:

- If the config is present, then don't parse the request body.
- If the config is present, then parse the body like we currently do.

Why:

Quick fix to allow this functionality in v2. We should've moved this to a service, but hindsight is always 20/20. We can move this to a service in v3. Issue is here: #668.

## Acceptance Criteria

- [ ] Write documentation
- [ ] Write unit tests
- [x] Write integration tests

crookse:
Closes #669 

## Summary

Adds an `optionals` server config that takes in a `request` optionals object. We can use this `optionals` config to introduce changes to different HTTP objects if needed.

Example:

```typescript
import * as Drash from "/path/to/drash/mod.ts";

class HomeResource extends Drash.Resource {
  public paths = [
    "/",
  ];

  public async POST(
    request: Drash.Request,
    response: Drash.Response,
  ): Promise<void> {
    // Since the body was not read, users can call the request methods themselves
    const body = await request.blob();

    console.log(body); // Example output => Blob { size: 15, type: "image/png" }

    response.text("Hello");
  }
}

  hostname: "0.0.0.0",
  port: 1447,
  protocol: "http",
  resources: [
    HomeResource,
  ],
  optionals: {
    request: {
      read_body: false, // Prevents the Drash.Request object from reading the body when it's created
    }
  }
});

server.run();

```

@ebebbington @Guergeiro, i initially had the `optionals` config as:

```js
optionals: {
  request: {
    content_types: [ ... ]
  }
}
```

but it seemed weird and added more code to maintain, so i made it `read_body`. imo it seems to be "fluent" like... "ok, yes this config means to read the body or not" instead of like "ok these are request content types... because why? lemme look at the documentation again"

### DIFF
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -72,6 +72,7 @@ export class DrashRequest extends Request {
    * @param pathParams - The path params to match the request's URL to. The path
    * params come from a resource's path(s).
    * @param connInfo - The connection info Deno provides on a new request
+   * @param requestOptions - Any options to control the way requests behave.
    *
    * @returns A Drash request object.
    */
@@ -79,7 +80,7 @@ export class DrashRequest extends Request {
     request: Request,
     pathParams: Map<string, string>,
     connInfo: ConnInfo,
-    requestOptions: RequestOptions,
+    requestOptions?: RequestOptions,
   ) {
     const req = new DrashRequest(request, pathParams, connInfo);
 
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -1,7 +1,7 @@
 import { getCookies } from "../../deps.ts";
 import { Errors } from "../../mod.ts";
 import type { ConnInfo } from "../../deps.ts";
-import { BodyFile, RequestOptionals } from "../types.ts";
+import { BodyFile, RequestOptions } from "../types.ts";
 
 export type ParsedBody =
   | Record<string, string | BodyFile | BodyFile[]>
@@ -79,12 +79,12 @@ export class DrashRequest extends Request {
     request: Request,
     pathParams: Map<string, string>,
     connInfo: ConnInfo,
-    optionals?: RequestOptionals,
+    requestOptions: RequestOptions,
   ) {
     const req = new DrashRequest(request, pathParams, connInfo);
 
-    if (optionals) {
-      if (optionals.read_body === false) {
+    if (requestOptions) {
+      if (requestOptions.read_body === false) {
         return req;
       }
     }
--- a/src/http/request.ts
+++ b/src/http/request.ts
@@ -72,7 +72,7 @@ export class DrashRequest extends Request {
    * @param pathParams - The path params to match the request's URL to. The path
    * params come from a resource's path(s).
    * @param connInfo - The connection info Deno provides on a new request
-   * @param requestOptions - Any options to control the way requests behave.
+   * @param options - Any options to control the way requests behave.
    *
    * @returns A Drash request object.
    */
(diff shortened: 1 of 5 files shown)


Guergeiro (approved):
Lgtm

ebebbington (approved):
i like it :) easy simple way for users to disable reading the body, gives more fine grained control

crookse:
@ebebbington @Guergeiro wondering if `optionals` should be something else because the server configs have optional configs. seems kind of confusing. you guys have an idea for a better name or is this one good?

Guergeiro:
Naming was, and still is, one of my pitfalls... It seems nice to me, but I get what you're saying. I don't have a solution unfortunately.

crookse:
> Nuke the optionals name, just have server.request.read body, I think we’re over complicating it, we’re specifying a request config on the server, this shows that we want to customise the request

doneski: https://github.com/drashland/drash/pull/671/commits/7e4c8613d3f4ab5db8be0452d691b7e8a5f3b082

src/http/request.ts
```diff
@@ -72,15 +72,24 @@ export class DrashRequest extends Request {
...
    * @param connInfo - The connection info Deno provides on a new request
+   * @param requestOptions - Any options to control the way requests behave.
    *
    * @returns A Drash request object.
    */
   static async create(
     request: Request,
     pathParams: Map<string, string>,
     connInfo: ConnInfo,
+    requestOptions?: RequestOptions,
```
saragee3, line 83: why not just `options`? 😛

src/http/request.ts
```diff
@@ -72,15 +72,24 @@ export class DrashRequest extends Request {
...
   ) {
     const req = new DrashRequest(request, pathParams, connInfo);
+
+    if (requestOptions) {
+      if (requestOptions.read_body === false) {
```
saragee3, line 88: `ignore_body` could be a good candidate... when i first read `read_body`, i thought it was a required bool while `ignore_body` conveys that it's an optional setting 🤔

src/http/request.ts
crookse: lmao probably coded too fast. the typing says request options and we're creating a request, so i think it should be `options` too

src/http/request.ts
crookse: i think this makes sense, but would we also make this change from the user's point of view? like would they pass in `ignore_body` when they create the server object?

saragee3: yeah, don't they have to do that now with `read_body`? currently, `read_body` defaults to `true` right? so i can see `ignore_body` defaults to `false`... but honestly no strong feelings ha

saragee3 (approved):
small comments but overall looks goooood

src/http/request.ts
crookse: doneski: https://github.com/drashland/drash/pull/671/commits/ec516042683eafafd6fbca0cfea1f59842e9b462

crookse:
## Usage

```typescript
  ...
  ...
  ...
    request: {
      read_body: false,
    }
});
```

PR #685, opened 2022-10-04
chore: Bump ResourceLoaderService to Deno Standard Modules 0.158.0

tweenietomatoes:
Bump `ResourceLoaderService` to Deno Standard Modules 0.158.0


crookse:
Thanks for the update @tweenietomatoes !

PR #690, opened 2022-12-16
fix: sending a POST request without a body

ebebbington:


ebebbington:
@crookse i only just realised we have `read_body` option, so instead of using original request, i thought we'd utilise that in our code, so people can pass that option in as false, and have full access to the body

As we only saved the original request purely for preserving the body, i also thought it made sense to only care about the body, no need to preverse the whole request

src/http/request.ts
```diff
@@ -56,7 +53,7 @@ export class DrashRequest extends Request {
   }
 
   get original(): Request {
-    return this.#original;
+    return Object.getPrototypeOf(this);
```
Guergeiro, line 56: Wait... I'm kind of lost. We can't clone the original request? Doesn't `Object.getPrototypeOf(this)` return the current instance of the request? So, if a request is changed, the prototype will also be changed, right?

crookse:
sorry for late review! i'll try to have a look at this tonight. it seems ok, but i'm surprised we needed to change the tests to accommodate the fix. thanks for looking into this tho!

crookse:
see #691 for another solution and to discuss (just adding this comment so we can link this PR and 691 together)

src/http/request.ts
crookse: we changed the approach in #691 @Guergeiro. i just merged #691 into this PR.

crookse (approved):

PR #691, opened 2022-12-26
fix: sending a POST request without a body

crookse:
@ebebbington, this approach passes in a request clone to `super()` in `DrashRequest`'s constructor. most of the current code is left intact, the old tests are being used, and there are also tests i added in addition to the tests you added.

when i reviewed your PR, i noticed we couldn't get the `request.original` without the code throwing `Illegal invocation` which means we'd break current functionality because we allow users to access `request.original`. see screenshot below:

Screen Shot 2022-12-25 at 21 00 24

i think the approach in this PR is the one we should go with. thoughts @ebebbington @Guergeiro? i could def be missing something here. this deno change seems like it had a big impact and we might just be hitting the surface with its effects.

### DIFF
--- a/tests/integration/post_no_body_test.ts
+++ b/tests/integration/post_no_body_test.ts
@@ -29,12 +29,11 @@ const server = new Server({
 Deno.test("post_no_body_test.ts", async (t) => {
   await t.step("POST /post-no-body", async (t) => {
     /**
-     * This bug was discovered by sending no body ina post request.
+     * See the following for reasons why this test was added:
      *
-     * The fix was to remove the `#original_request` on `request.ts` as we cloned it,
-     * no body seemed to not be caught by the native code, but we didn't need it anyway.
+     *   - https://github.com/drashland/drash/pull/691
      */
-    await t.step("Does not throw if a body isnt defined", async () => {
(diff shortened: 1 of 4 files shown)


tests/integration/post_no_body_test.ts
```diff
@@ -0,0 +1,51 @@
...
+////////////////////////////////////////////////////////////////////////////////
+
+Deno.test("post_no_body_test.ts", async (t) => {
+  await t.step("POST /post-no-body", async (t) => {
+    /**
+     * This bug was discovered by sending no body ina post request.
+     *
+     * The fix was to remove the `#original_request` on `request.ts` as we cloned it,
+     * no body seemed to not be caught by the native code, but we didn't need it anyway.
+     */
```
ebebbington, line 36: my comments probs redundent now lol

ebebbington (approved):
niceee much better

tests/integration/post_no_body_test.ts
crookse: removed! [952800c](https://github.com/drashland/drash/pull/691/commits/952800c9e75a402c38c1428fc9965002806ed4a3). actually just changed it to reference this PR.

